PR #9, opened 2021-01-06
Updated Readme.md - Added alternativ lineage name

corneliusroemer:
To keep section consistent, added alternative lineage name **B.1.351** to the South African variant, since the English variant also has **B.1.1.7** in parentheses
For usage of this term, see e.g. here https://cov-lineages.org/global_report_B.1.351.html


emmahodcroft:
Thanks!

PR #15, opened 2021-01-19
UI improvements in variant display

jameshadfield:
Makes two changes:

1. Groups the sidebar into different categories which will help distinguish between different "things" we hear about - clades, lineages, single mutations etc
2. Removes the "defining mutations" header from pages which didn't have any defining mutations (e.g. single spike mutations). Having this heading was confusing as the markdown content went directly below it.

Note: I'm not great at typescript (yet) ~so there are some typescript errors here. Sorry!~ so would appreciate comments here relating to this!

cc @emmahodcroft @ivan-aksamentov


ivan-aksamentov:
Ha, cool! Thanks James. Also, quick digging into the code. Impressive!


It's up to Emma to make the decisions about features (usefulness, scientific soundness etc.), she's the boss here, but the code looks good, including Typescript. 

There is only [one warning from DeepScan](https://deepscan.io/dashboard/#view=project&tid=12386&pid=15414&bid=307276&subview=pull-request&prid=614549) about missing React key prop. Might screw the rendering a bit.

Optionally, stylistically:
 - We could also make this `<>` Fragment a `div`, or maybe even a `Card`? Although that might also introduce too many borders and outlines and waste space.

 - Would be great to center the headings on mobile.


Emma is going to announce on Twitter, so the first visit will be pretty much always from mobile. We tried to get it right on Phone sizes.

This is iPhone X as Chromium device toolbar shows:

<a target="_blank" href="https://user-images.githubusercontent.com/9403403/104995928-50c14580-5a27-11eb-82e6-0c41a8dad1e5.png">

</a>

<a target="_blank" href="https://user-images.githubusercontent.com/9403403/104995935-53bc3600-5a27-11eb-864e-f715bbceeda0.png">

</a>

"Defining mutations" are also super ugly still... We might also try and group them separately.


Note to Emma: you can review the app changes by using the "Preview" link in Vercel bot's message:
https://github.com/hodcroftlab/covariants/pull/15#issuecomment-762615883

emmahodcroft:
@jameshadfield I really like this idea of diving up the menu, and I think I do want to implement this. My only thought right now is that thanks to the fact this has grown organically, it's a bit tricky to know where this divide might be. I am planning in the near future to split this up a bit more so that the key variants (501Y.V1/2/3) have their own buttons - then they'll fit in better with the EU1, EU2, etc, and I think this'll make the divide more clear. So I'll probably sit on this until I get that split-out done, then pull this back in!

emmahodcroft:
The PR #115 should take the last suggestion from this branch into effect - splitting out the menu bar! 
Sorry this took so long to implement James, but I'm glad we've got it in there at last!

Issue #26, opened 2021-01-19
Tweak mutations colors for better readability


ivan-aksamentov:
Some of the colors on badges in "Defining mutations" section have insufficient contrast ratio against text and/or background. Simply put, it's hard to read the text.

We want to change the colors, for better readability.

The colors are here
https://github.com/hodcroftlab/covariants/blob/master/web/src/colors.ts

Originally, the Nuc and AA colors are taken from bioSyntax project:
https://github.com/bioSyntax/bioSyntax
more specifically, their spreadsheet
https://github.com/bioSyntax/bioSyntax/blob/429671cfbf6d7355e1db25dc8116687dd8683029/dev/theme/bioSyntax_theme.xls

The gene colors in the same file are taken from Nextclade, which are taken from Auspice. These are the same as in the Auspice entropy panel.

That is, there are less colors in auspice than we need and auspice reuses the same colors in cycle. I ran this cycle for SARS-CoV-2 genes and made a harcoded dictionary. I also made gene S to be red and gene N to be blue, don't remember exactly, so that they are not of the same color.
https://github.com/nextstrain/nextclade/blob/1fbc1e924bd3f58e0681f0430d033e5dee83e776/packages/web/src/constants.ts#L31-L44

Additional points for making the colors colorbindness-friendly, but given how many colors there are, it's a tough task.

emmahodcroft:
One idea posed by Ivan: make colours slightly more pastel so they are a bit less glaring. But will preserve seeing similar AA have similar properties :)

emmahodcroft:
A nice colour scheme found by Richard: http://ugene.net/forum/YaBB.pl?num=1337064665 (would still perhaps need a little tweaking), with good explanations of the colours, too.

ivan-aksamentov:
We could also add somewhere a legend that will explain colors, if that's useful.

emmahodcroft:
I was thinking this might fit in well in the FAQ section perhaps?

PR #40, opened 2021-01-19
Tweak mutations colors for better readability

ivan-aksamentov:
The aminoacid colors are taken from 
http://ugene.net/forum/YaBB.pl?num=1337064665

All colors are made dimmer and more pastel, to facilitate reading.
Might need a little more tweaking before it's perfect.

Resolves: #26


emmahodcroft:
From a quick glance, I like this *a lot*. I will try to take a closer look, using a wider range of AAs to be sure we like how they look in 'real life'. My only comment thus far is that the `T` in Amino Acid may blend in a bit too much with the grey that's the color for the position:
image

However this may be highly monitor dependent...

ivan-aksamentov (commented):
Made T a bit more saturated

emmahodcroft:
Yes! I think that's great!

emmahodcroft:
here's a demo with all AAs and all genes:
image

I think it's looking good!

emmahodcroft:
I think perhaps the only one left I might edit is `B` - as it looks a lot like the `-` `X` `*` symbols.
I'll take a look.

emmahodcroft:
Unless there are remaining thoughts about the colours or anything else, I'd be happy to merge this now!

ivan-aksamentov:
LGTM. Definitive merge

PR #47, opened 2021-01-20
Only display "defining mutations" header if there are defining mutations

ivan-aksamentov:
Cherry-picks 562ac7512d939d844b7cf9861acf45fd85fed0bb by @jameshadfield 

Hides "Defining mutations" section is there are no defining mutations


emmahodcroft:
I like this commit - thanks @jameshadfield ! I have added markdown headings called `## Mutation Information` to all pages where they don't have defining mutations, so that this shows up in a sensible way.

However - there's a bug somewhere in this code - for `S:V1122` and `S:VA626` there are no mutations displayed:
image


Should be like this (from master):
image


I imagine this is a bug in the boolean logic for when to display the heading. These two are the only two mutations that only have entries for `nonsynonymous` in the `clusters.py` file - instead of either having neither, or having both `nonsynonymous` and `synonymous` - so somehow it's not being picked up!

ivan-aksamentov:
> bug somewhere in this code - for S:V1122 and S:VA626 there are no mutations displayed
should be resolved in 2c12548

emmahodcroft:
This seems fixed to me now! happy to merge

ivan-aksamentov:
@emmahodcroft LGTM

jameshadfield:
\> bug somewhere in this code - for S:V1122 and S:VA626 there are no mutations displayed

image

PR #49, opened 2021-01-21
fix: use data from clusters.py instead of hardcoded for urls

ivan-aksamentov:
Cluster entries in `clusters.py` now contain `nextstrain_url` property, which contains full URL to Nextstrain build.

The URL construction logic is now, hopefully, removed from all paces.


ivan-aksamentov:
@emmahodcroft the app should now use `nextstrain_url`. It seems to be working, but please double-check the that the links are indeed scientifically sound.

emmahodcroft:
Have manually checked all the links in the deployment - and all look good!
If this is otherwise ready to go with the code on your end, we can merge!

PR #52, opened 2021-01-22
Add mutation comparison

ivan-aksamentov:


emmahodcroft:
This is looking super, super cool Ivan!!

emmahodcroft:
One note - can we change from 'individual mutations' to 'Other mutations' ?  - I think here it's not quite the same, and people might think individual means the same as in NextClade, which it isn't here.

ivan-aksamentov:
> change from 'individual mutations' to 'Other mutations' 

@emmahodcroft done!

PR #60, opened 2021-01-27
feat: add border to variant badges

ivan-aksamentov:
Depends on #59 merge it first!

optionally we could change the color of the S gene to make it more distinct from the link border color


emmahodcroft:
Neither @ivan-aksamentov or I are satisfied that this really makes the badges look like links. Closing for now in search of a better solution/idea.

PR #63, opened 2021-01-28
Update S.N501.md  ORF1a(Nsp6) deletion in 501Y.v1-2-3,

acx01b:
Just mentioning the striking ORF1a(Nsp6) deletion in 501Y.v1-2-3, because I like it a lot, 
I am probably not aware of the latest developments so maybe you don't want to mention it


emmahodcroft:
Thanks! Yes, definitely worth mentioning this mutation! I'll copy these over to the 484 page, which has much of the same text for V2 and V3 👍

PR #64, opened 2021-01-28
Fix Githuburl link from a variant page to point to build_name

### LINKED ISSUE #65: minor: "Propose changes to this section" for Variants: githubUrl needs ':' --> '.'
First, beautiful work on this, Emma and Ivan!  I just noticed a minor thing.  On variant pages like https://covariants.org/variants/S.N501 , the "Propose changes to this section" link is broken because it uses the variant name with a colon ([S:501](https://github.com/hodcroftlab/covariants/blob/master/content/clusters/S:N501.md), 404 not found) instead of a dot ([S.501](https://github.com/hodcroftlab/covariants/blob/master/content/clusters/S.N501.md)).

A quick look at the nice clean code implicates [this line](https://github.com/hodcroftlab/covariants/blob/ab13d3b75dbc233033a141dfa7b1c9fddd55580b/web/src/components/Variants/VariantsPage.tsx#L83) (and here my knowledge of your system ends so I have no suggestion for how to fix it, but I know you will!).

theosanderson:
instead of `display_name`

Otherwise you get a 404.

Tested locally

Closes #65

### DIFF
--- a/web/src/components/Variants/VariantsPage.tsx
+++ b/web/src/components/Variants/VariantsPage.tsx
@@ -80,7 +80,7 @@ export function VariantsPageDisconnected({ currentCluster }: VariantsPageProps) 
           </Col>
 
           <Col lg={9} xl={10}>
-            <EditableClusterContent githubUrl={`blob/master/content/clusters/${currentCluster.display_name}.md`}>
+            <EditableClusterContent githubUrl={`blob/master/content/clusters/${currentCluster.build_name}.md`}>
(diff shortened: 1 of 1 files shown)


theosanderson:
(Looks like the Vercel preview isn't working right now - I'll let you know when I test that the change doesn't break things)

emmahodcroft:
Thanks @theosanderson ! We are still waiting for the Vercel support to come through...! 🙃 Hopefully the auto-deploys will work pretty soon!

PR #66, opened 2021-01-28
Use build_name not display_name for image files

### LINKED ISSUE #18: Use build_name  for filenames for consistency
We transited to `cluster.build_name` (as in `scripts/clusters.py`) for content (`.md`) filenames (see https://github.com/hodcroftlab/covariants/commit/f76f6aa9ec443108109dd2b668204121bfd75a29), but other places might still use `display_name`. 

Notably `.gif` images and `.jpg` stulls created from them are using `display_name`.

We might change that for:
 - consistency (confusing and error prone when adding new content)
 - better compat with Windows (which apparently still has problems with colons in filenames in 2021)
by using `build_name` instead.

theosanderson:
`display_name` contains colons, which requires the filenames it points to contain colons, which mean the repo cannot be cloned on Windows. This changes image files to be pointed to by a filename including `build_name` in place of `display_name`, and renames the files to match up.

Fixes #18 

(Tested that images still appear with a local version)


ivan-aksamentov:
Thanks @theosanderson !

PR #71, opened 2021-01-29
Add basic windows dev support

theosanderson:
This adds hardcoding to a path that seems to work in Windows. (See discussion in https://github.com/nextstrain/nextclade/issues/313 and https://github.com/hodcroftlab/covariants/issues/18).

It's a bit hacky, I don't have a good understanding of structuring node commands. Improvements welcome.

### DIFF
--- a/docs/Contributing.md
+++ b/docs/Contributing.md
@@ -15,8 +15,18 @@ Navigate to `covariants/web` and type the folowing commands:
 ```
 cp .env.example .env
 nvm use
-yarn dev
 ```
+
+then start the development server with:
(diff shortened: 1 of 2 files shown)


theosanderson:
It occurs to me that the `_win` commands here should? work for Unix too? Maybe we can just edit the existing commands to use this alternative path.

ivan-aksamentov:
Thanks Theo! I will merge as is. These shell scripts probably still have some use. Let's keep the Unix npm scripts ones as they are, to avoid possible breakage.

PR #75, opened 2021-02-01
feat: add 'total' and 'others' to per-country plot tooltips

### LINKED ISSUE #72: Add 'Other' Count to Tooltip in 'Per Country' graphs.
When mousing over 'Per Country' graphs, the tool tip shows the sequence count for each cluster - this is very useful! However, it does not show the 'other' (the white space which doesn't fall into any cluster).
image

Adding a column to the tooltip called 'other' and having a count would be useful for users to estimate how many sequences did not have a cluster assigned, and how many sequences there were in total (adding up all those with a cluster + those without). This would be particularly helpful if only one cluster is visible (through changing the check-boxes on the left):
image

This needs to be dynamic as changing the clusters that are visible will impact how much 'white'/'other' there is - so needs to be calculated 'on-the-fly'. 

In the file used to generate these charts we do already know the 'total' for each time point and the number of sequence in each cluster, so the 'other' is simply (`total` - `number of sequences in visible clusters`).

ivan-aksamentov:
Resolves #72 

Adds 2 rows:
 - "total" number of sequences at the bottom
 - "others" as a difference of total and sum of number of sequences for all variants (variants which are not filtered out with checkboxes)

Related: #76

### DIFF
--- a/web/src/components/CountryDistribution/CountryDistributionPlot.tsx
+++ b/web/src/components/CountryDistribution/CountryDistributionPlot.tsx
@@ -34,7 +34,7 @@ export function CountryDistributionPlotComponent({ cluster_names, distribution }
 
     const others = total_sequences - total_cluster_sequences
     const weekSec = DateTime.fromFormat(week, 'yyyy-MM-dd').toSeconds()
-    return { week: weekSec, ...cluster_counts, others }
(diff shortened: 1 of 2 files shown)


emmahodcroft:
This is great Ivan!! Thank you!

PR #76, opened 2021-02-01
feat: add frequency column for per-country plot tooltips

ivan-aksamentov:
Extends and depends on #75

Note: This PR is done, but is marked as draft, to avoid out-of order merges, because it requires #75 to be merged first. Safe to unmark (click "Ready for review") and merge when #75  is on master.

This adds another column "Freq", which value is `number_of_sequences / total_sequences`


emmahodcroft:
This is also super-great @ivan-aksamentov !!

PR #84, opened 2021-02-07
Update Home.md

under-score:
deep link "headline" names


emmahodcroft:
Thanks for this suggestion - we've implemented something similar now in the badges + names table on the home page!

PR #86, opened 2021-02-08
feat: add color to "others" variant on per-country plots

### LINKED ISSUE #85: Consider making `others` area grey not white in per-country plots
Small suggestion. In my view the per-country area plots are a bit hard to read visually because it's not clear that the white space represents also represents SARS-CoV2 cases (which makes it seem like it's not a plot of proportions). I think making the "other" slice appear in grey so it is clearly a defined proportion might help with this.

ivan-aksamentov:
Resolves #85

This adds fill color to "others" entry on the per-country plots. The color is reused in the plot tooltip, to fill the rectangle near the "others" entry as well.

The color can be adjusted here: 
https://github.com/hodcroftlab/covariants/blob/2f6483d978a500f8302bdd55a9a1758252a69a66/web/src/theme.ts#L149


emmahodcroft:
This looks great @ivan-aksamentov! Thanks for tackling this! I think this is good, with just a request for a slightly lighter shade of grey, maybe more like `#cccccc` ?

emmahodcroft:
This looks pretty good to me now, if I haven't messed anything up. We could try a slightly lighter grey still, but dunno.

ivan-aksamentov:
LGTM as is

Issue #80, opened 2021-02-02
Avoid duplicated legend hints across countries


ivan-aksamentov:
The issue is raised in this comment: https://github.com/hodcroftlab/covariants/pull/77#issuecomment-772014478

If two countries have the same color they are indistiguishable in tooltips and filtering sidebar. They can be distinguished in plots, because the lines have different styles (e.g. solid vs dashed).

We could solve this by drawing short pieces of lines instead of circles in tooltips and filtering panel.

Issue #81, opened 2021-02-02
Automate country color generation


ivan-aksamentov:
We could make a variety of colors by generating them automatically, given country names. There are packages which allow to "hash" strings into colors, producing an (almost) unique color given a string, for example
https://pypi.org/project/colorhash/

Color hashes are stable, that is, they produce the same color if given the same string.

This can produce broader variety. But also does not guarantee that colors will be easily distinguishable from each other. We might tweak them individually, by adding a "tint" to a given country, or by adding a common "salt" or an individual "pepper" to the hashed string, or by combining all.

Individual tint:

```
color = ColorHash('France')
color = add_tint(color, countries['France].tint)
```

Common salt:
```
color_fr = ColorHash('France' + salt)
color_de = ColorHash('Germany' + salt)
```

Individual pepper:
```
color_fr = ColorHash('France' + countries['France].pepper)
color_de = ColorHash('Germany' + countries['Germany].pepper)
```

Related: #80

emmahodcroft:
I like this idea as a long-term solution!

PR #87, opened 2021-02-08
feat: use styled lines instead of colored circles on "per variant" page

ivan-aksamentov:
Resolves https://github.com/hodcroftlab/covariants/issues/80

This replaces colored circles with short colored dashed lines as icons for country entries in plot tooltips and in filtering sidebar on "per variant" page. This way the countries with the same color look distinct if they have different line styles. The colors and styles mimic those in the plot.

The issue of duplicated colors still persists on "per cluster" page, which still uses circles. Using dashed lines there makes less sense, as there are not line plots on that page. Let's discuss this issue separately.

Additionally, some colors were too pale even when circles were used. And now with lines, especially dashed, these elements are almost indistinguishable from the background. This is poor user experience, especially in terms of accessibility.

Related: https://github.com/hodcroftlab/covariants/issues/81


emmahodcroft:
I like this, thanks Ivan - this is super good. I agree that we can improve this through better colour choice and line style. It might be worth having an offline discussion about how we can mix-and-match these to avoid light, dashed colours!

We seem to have one conflict but I'm happy to merge after that's resolved.

ivan-aksamentov:
@emmahodcroft merge conflict should be resolved now. Ready to go.

emmahodcroft:
Thanks Ivan, this is great!

PR #89, opened 2021-02-08
feat: modify GISAID acknowledgements

ivan-aksamentov:
Modifies mentions of GISAID to adhere to their guidelines:

 - [x] Moves "Enabled by data from GISAID" text to the top
 - [x] Adds "GISAID data provided on this website are a subject to GISAID Terms and Conditions" text linking to https://www.gisaid.org/registration/terms-of-use/ (URL correctness is to be verified)
 - [x] Removes GISAID logo from "Powered By"  section of the footer
 - [x] Adds "Acknowledgements" dialog


emmahodcroft:
This is looking great @ivan-aksamentov - thank you for working on this! I noticed that the footer is now a stack of 3 lines:
image

rather than spaced on one line, left, middle, right (with 'middle' added to our existing left and right):
image

I don't have strong feelings here so if that's easier, no problem. I just wanted to check if it's intentional?

ivan-aksamentov:
@emmahodcroft 

> I just wanted to check if it's intentional?

It is. These strings now only fit into one line on very large screens. I did not bother and put them stacked.

emmahodcroft:
Ivan I noticed when I click 'Acknowledgements' then close it, there's a black bar around the button that remains:
image

Also, for an idea to fix the broken iPad view, could we make the menu wrap? So that acknowledgements goes below 'Home'? Just an idea, though you may have a better one. When we have something we think is good I can send GISAID a deployed link to see if they think all is ok.

emmahodcroft:
Merging in to get these GISAID changes live!

PR #92, opened 2021-02-12
make plot titles on per-variant page links

jamesscottbrown:
This makes the plot titles on the "Per variant" page links to the page for the corresponding variant.


ivan-aksamentov:
Thanks @jamesscottbrown,

I'll comment on the dev part, but it's Emma's @emmahodcroft call whether to consider this feature or not.

Let's not use `cluster.disply_name.replace()`. It will not work in general case. Instead we should use `cluster.build_name` (coming from [data](https://github.com/hodcroftlab/covariants/blob/d984abf4a81a1e82e4929344a02bbb32f828ec3c/web/data/clusters.json#L4)), as a single source of truth for URLs.

Personally, I am not a huge fan of the large blue headings. Maybe we could style them to remove the color and underlining?

jamesscottbrown:
> Let's not use cluster.disply_name.replace(). It will not work in general case. Instead we should use cluster.build_name (coming from data), as a single source of truth for URLs.

I've updated this.

> Personally, I am not a huge fan of the large blue headings. Maybe we could style them to remove the color and underlining?

The underlining only appears whilst the cursor is hovering over the link.

I've changed the link color back to the original grey - I agree that this looks better, though it is less obvious that they are links before they are moused-over.

emmahodcroft:
I like this PR. @ivan-aksamentov If the updates take care of your issues above, I'm happy to merge. I agree the reverting to the original style means the links aren't as visible, but I agree with Ivan that the 'big blue' didn't look ideal either. I'm happy to work on making this more visible in another way, but this can be by other PRs - the links can still work in the meantime for those who discover them :)

web/src/components/ClusterDistribution/ClusterDistributionPage.tsx
```diff
@@ -28,6 +29,8 @@ const CLUSTERS_STATE = CLUSTERS.reduce((result, cluster) => {
   return { ...result, [cluster]: { enabled: true } }
 }, {})
 
+const CLUSTER_BUILD_NAMES: Map<string, string> = new Map(getClusters().map((c) => [c.display_name, c.build_name]))
+
```
ivan-aksamentov, line 33: Instead of this, could we pass the whole `cluster` object from parent (and grandparent, grandgrandparent,... if necessary), and use `cluster.build_name` and `cluster.build_name`? I'm afraid current code might trigger unnecessary re-renders.

web/src/components/ClusterDistribution/ClusterDistributionPlotCard.tsx
```diff
@@ -2,28 +2,41 @@
...
+  const url = `/variants/${clusterBuildName}`
+
   return (
     <Card className="m-2">
       <CardHeader className="d-flex flex-sm-column">
         <PlotCardTitle>
-          <span>{cluster}</span>
+          <Link href={url}>
+            <GreyLink>{clusterDisplayName}</GreyLink>
+          </Link>
```
ivan-aksamentov, line 39: `Link` component already contains `a` inside, so this will generate incorrect HTML. Instead, the styling should be applied to the `Link` component directly.

Also, it is preferable to use the function syntax for retrieving theme properties, instead of importing. This way themes can be swapped on runtime.

Example:
https://github.com/hodcroftlab/covariants/blob/cf0283b57f3626adf9513ea506bd93949569a46a/web/src/components/ClusterButtonPanel/ClusterButton.tsx#L41-L71

ivan-aksamentov (approved):
Added a couple of comments.

@jamesscottbrown please take a look if you have a second. Otherwise I can try, but a bit later. Currently busy on another project.

web/src/components/ClusterDistribution/ClusterDistributionPage.tsx
ivan-aksamentov: This is tricky due to filtering logic in the component, so let's keep it this way for now and address it Later™ :)

web/src/components/ClusterDistribution/ClusterDistributionPlotCard.tsx
ivan-aksamentov: Fixed this in [`f1c8a00` (#92)](https://github.com/hodcroftlab/covariants/pull/92/commits/f1c8a000b7589a485b3e082b36ad67e5d962acab)

ivan-aksamentov:
Addressed the redundant `<a>` element part.
Created an issue for the build_name workaround to be fixed later: https://github.com/hodcroftlab/covariants/issues/100

PR #94, opened 2021-02-12
add tooltip text to mutation badges

jamesscottbrown:
This adds an explanation of the mutation badge as a `title` attribute, so that it appears on mouse-hover.

The screenshots illustrate this, but with the cursor hidden:

non_synonymous

synonymous


ivan-aksamentov:
Thanks @jamesscottbrown ,

I'll comment on the dev part, but it's Emma's @emmahodcroft call whether to consider this feature or not.

I guess it does not hurt anyone. My goal is to ensure nothing blows up :)

 1. Could you please double check all the object lookups? I can never remember if it throws an exception or returns undefined if a key is not found. I'd rather use `lodash` `get(obj, key, defaultValue)` instead of plain lookups like `AMINOACID_NAMES[left]` (we depend on lodash anyways already). This should also simplify the code a lot, eliminating all the conditionals, including the one with the `gene` variable.

 2. We might clarify the var names. What is `wt`? I used plain silly `left` and `right` to not make any assumptions. Also we use camelCase convention.

 3. If Emma approves, For symmetry maybe it also makes sense to also add tooltips to the variant badges? Although with variant naming conventions being a complete mess it might be tricky and annoying to maintain.

What do you think?

emmahodcroft:
@jamesscottbrown Thank you for this PR! This is certainly an interesting idea and one that could help clarify things for users. I'll try to test this out myself and see how it looks soon!

jamesscottbrown:
> Could you please double check all the object lookups? I can never remember if it throws an exception or returns undefined if a key is not found. I'd rather use lodash get(obj, key, defaultValue) instead of plain lookups like AMINOACID_NAMES[left] (we depend on lodash anyways already). This should also simplify the code a lot, eliminating all the conditionals, including the one with the gene variable.

It returns undefined rather than throwing an exception, but since the project is using lodash using `get` is better - I've changed this.

>  We might clarify the var names. What is wt? I used plain silly left and right to not make any assumptions. 

`wt` here is Wild Type (i.e., not mutant or variant). This seemed clearer than left or right, as it reflects what it is rather than where it is written in the badge. I've expanded the name, as `wildTypeBase` is still quite short.

> Also we use camelCase convention.

Sorry - I had been writing Python all day.

ivan-aksamentov:
Looks okay dev-wise. Let's see what science department says :)

emmahodcroft:
Thank you for doing this @jamesscottbrown - I think this is really helpful after trying it out. I can never remember the amino acids myself, and I imagine I'm not the only one, so this is hopefully very useful to others as well! @ivan-aksamentov if this is all 👍 from your end, we can merge!

PR #99, opened 2021-02-17
Tooltip sorting semantics

jamesscottbrown:
The tooltip sorting currently offers a choice of sorting by frequency or country, and by 'reversed' or not.

The non-reversed order is currently ascending order for both frequency and country name. 

The default selection is "frequency" and "reversed", which is the most useful ordering; but changing from "frequency" to "country" results in reverse-alphabetical ordering, which is unexpected.

I think it is would make more sense for the non-reversed ordering to be descending for frequency and ascending for country name. This would mean that the "reversed" checkbox is not initially checked.


ivan-aksamentov:
Thanks!

Yes, I think I screwed up the UX here. Later on we might just have 4 dropdown options {freq, coutry} x {ascending, descending}, like on most of the shopping websites. We might also add sorting for the clusters page as well, for symmetry.

LGTM, Emma feel free to merge.

ivan-aksamentov (approved):

emmahodcroft:
This is great. Thanks to you both! 🎉

PR #102, opened 2021-02-18
Highlight line when country name moused-over

jamesscottbrown:
One way to address the problem of distinguishing between many different lines is with interactivity.

I tried implementing a features that highlights the corresponding line on the plot when a user mouses-over a country name in the panel on the left.
A user can thus rapidly scrub their cursor down the list of names, and see the trend for that country in the plots for each variant.

When mousing-over Spain (the cursor does not appear in the screenshot), this looks like:

spain

However, performance is unacceptably bad, and there is a noticeable delay between moving the cursor to another country name and the plots updating.

The flamegraph in React dev tools suggest that re-rendering each line-chart takes roughly 50-60ms, even though only two lines have changed thickness and need to be re-rendered, and the others are unaffected.

flamegraph

I haven't used recharts before and I'm not sure what's causing this problem or whether there's an easy solution.


ivan-aksamentov:
Wow thanks James @jamesscottbrown, this is cool!

Indeed Recharts are horribly slow. Not sure how to overcome that.

I noticed that lots of stuff is getting rerendered. Same on live site clicking a checkbox. So investigation is due here.

EPILEPSY WARNING: blinking content below

<details>
<summary>
Rerendering when clicking a checkbox (click to open)
</summary>



</details>

emmahodcroft:
Hey @jamesscottbrown - this is awesome! I love this as a feature and can see how it would be very useful indeed! 

It seems like we're going to need to tackle the fact that so much is getting rerendered too often in order to have smoother performance, but I hope that once we get that in place we can totally implement this!

PR #117, opened 2021-03-04
Per variant lines

emmahodcroft:
Allow more diverse colours and line patterns to help distinguish different countries on the per-variant page.


emmahodcroft:
Merging in because I keep running out of colors!

PR #118, opened 2021-03-05
color code in link to match rest of link

jamesscottbrown:
Currently the links to specific nextstrain builds look like two links, because the names formatted with `<code>` are in black rather than blue:

Currently:
old

After this PR:
new


emmahodcroft:
Thanks @jamesscottbrown ! This looks great to me, and I agree - it's much clearer! @ivan-aksamentov is on a much-deserved break, but I'll let him look this over when he's back, and then I think we can merge!

PR #121, opened 2021-03-06
Fix filtering in link to 20B/S.484K

theosanderson:
The link to the build here appears to neither filter nor zoom correctly. This I think fixes the filtering, but does not fix the zooming. Both are ticketed in https://github.com/hodcroftlab/covariants/issues/122


emmahodcroft:
Thank you for this @theosanderson ! I replied to #122 - in short, I switched the way the clades are called in the analysis but didn't update the URLS.

However, the `:` symbol that I use now doesn't actually work for all the URL functionality, so I'll go back to using `.` - I hope to update this in the next few days, and hopefully the URLs will work again! I'll leave this PR and the issue open until fixed to help me remember to check :)

theosanderson:
Fab, do close as soon as you like!

emmahodcroft:
Fixed now!

PR #125, opened 2021-03-11
Ensure variant badges work for named variants

ivan-aksamentov:
This:
 - makes sure that badges for variants "20E (EU1)" and "20A.EU2" are rendered correctly
 - adds unit tests for variant string parser for some of the variants
Screenshot from 2021-03-11 16-24-15


emmahodcroft:
This looks fantastic Ivan, thank you!

PR #126, opened 2021-03-11
Add lineage badges

ivan-aksamentov:
This adds badges for Pango Lineages

To be used as so:

```ts
<Lin name="B1.1.7" />
```

This will automatically link to a report:

```ts
<Lin name="B1.1.7" report />
```

This will link to a custom URL:

```ts
<Lin name="B1.1.7" href="https://example.com" />
```

The colors are generated using a color hash algorithm: https://github.com/hodcroftlab/covariants/blob/aadeb941048999ae8ccffc9a1e0f4934a9ca302a/web/src/components/Common/MutationBadge.tsx#L290-L297

The `colorHash()` function takes an arbitrary string (in this case, a sub-component in the name of the lineage) and returns a hex color string, which is specific for the original string. The colors are not guaranteed to be unique, but they are repeatable, that is the same input string generates the same color every time. There are parameters, which can be tweaked to modify colors: for example `lighntess` defines how dark are the colors, and `reverse` says whether to reverse the string before calculating the hash, in which case similar strings result in more distinct colors, etc.


emmahodcroft:
I really like this @ivan-aksamentov ! My only concern is that it is harder to tell which lineages link out to a report. I know we've dealt with this before and it wasn't an easy solution. Any chance we could just add the 'link out' icon after them, if they have a link? That might be easiest (and since I don't plan on using them elsewhere on the site, it shouldn't interrupt the flow of text). But happy to consider other ideas!

emmahodcroft:
Also, I don't know if this would be part of this PR or not, but I was hoping perhaps we could modify how the table shows in markdown - so that it draws vertical lines, and maybe a horizontal one under the 'heading'? I'm adding those lines in by just using a `|` symbol at the moment but it's not ideal as they aren't connected! 😆

emmahodcroft:
I hope I didn't mess up git history by fixing the conflict, I just thought maybe it was better to do it sooner rather than later? 😬

PR #128, opened 2021-03-11
Allow clusters to not have a nextstrain link

### LINKED ISSUE #119: Allow pages to not have a Nextstrain link
To finish breaking out the 'mutations' from the variants, I'd like to move the last of the mutation information out of the variant pages and into their own mutation pages (then import it to the variant pages).

However, I probably won't have a build for every mutation that has a page. It would be nice to have an option that if there's no Nextstrain URL (or perhaps, if it's got the value `False`) then instead of showing the Nextstrain build link at the top, it will just say 'No dedicated Nextstrain build is available' or similar.

ivan-aksamentov:
If a cluster entry in `clusters.py` does not have a `nextstrain_url` field, then Nextstrain link will not be shown on cluster's page. The text "No dedicated Nextstrain build is available" will be show instead.

Resolves #119


emmahodcroft:
Brilliant Ivan!! 😁

PR #133, opened 2021-03-17
don't add so many labels to new issues by default

jamesscottbrown:
The `help wanted` and `good first issue` tags should probably be added (if appropriate) when an issue is triaged, rather than being added immediately by default.

### DIFF
--- a/.github/ISSUE_TEMPLATE/01_report_bug.md
+++ b/.github/ISSUE_TEMPLATE/01_report_bug.md
@@ -1,5 +1,5 @@
 ---
 name: 🐛 Report a bug
 about: If something is broken
-labels: bug, help wanted, good first issue, needs triage
+labels: bug, needs triage
 ---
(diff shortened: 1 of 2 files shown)


emmahodcroft:
@ivan-aksamentov Could you look into this? It's probably true that not all issue are good first issues by default, though I suppose we often by default would like "help" 😆

jamesscottbrown:
> I suppose we often by default would like "help"

I suspect this will usually be the case, but probably not always:there might be some well-intentioned suggestions which you or Ivan think are actually bad ideas, so don't want people to help implementing, or which you want to be implemented in a very specific way.

ivan-aksamentov:
@jamesscottbrown Thanks for the suggestion. Internally we decided to keep both labels by default. This is for reasons unrelated to the real meaning of these labels - for better visibility in GitHub search and GitHub Explore, to increase discoverability and reach. Which is a bit spammy, but I would write it off this time, because it's an "emergency" pandemic software and also relies on external content contributors.

The triage suggestion would not work, as we often forget not only to add relevant labels, but even to remove the "needs triage" label (there are closed issues with this label) and this procedure is not enforceable.

Issue #88, opened 2021-02-08
Idea: use country flags instead of colored circles in legends


ivan-aksamentov:
Related: #81 #87

The idea is to use country flag icons instead of colored circles on "Per country" page for plot headings and filtering sidebar.

Originally, the country colors were meant for line plots in "Per variant" page, and they have no significance for the "Per country" page. But it was making sense to also use the same colors in country icons to connect the two pages together.

Now that there are more countries added, some colors started to be duplicated. On "Per variant" page the logical solution was to reuse the line styles to make the countries visually distinct #87. However, it might be confusing to use the same kind of line icons on "Per country" page, which does not have line plots.

This should solve the problem of duplicated colors on "Per country" page.

emmahodcroft:
I agree with you Ivan, that this makes more sense than the colours! I'd be happy if this implemented, though I'd say it's not high priority.

PR #135, opened 2021-03-25
replaced circles with flags

jamietyger:
### Idea: use country flags instead of colored circles in legends #88 

Added emoji's to the build to replace the colored circles

**web/data/countryStyles.json** 

Added flag emoji's for each country in dict

 **web/src/io/getCountryColor.ts**

Added getCountryFlag function and updated default record

 **web/src/components/DistributionSidebar/CountryFilters.tsx** 

Added flag emoji to the sidebar filter

 **web/src/components/CountryDistribution/CountryDistributionPlotCard.tsx** 

Added flag emoji next to the country name on chart


ivan-aksamentov:
Hi Jamie @jamietyger,

This looks cool!

Could you please:
 - extract the fallback icon into a global variable and silence the "ascii only" eslint error for this line.
 - fix remaining linter warnings and formatting problems (use `yarn format` and `yarn lint` commands locally, and/or see the results of DeepScan and TravisCI checks above)
 - revert unrelated changes to `package.json` and `yarn.lock`
 - remove `package-lock.json` (we use yarn as a package manager)
?

I will let Emma @emmahodcroft to review the usability/looks.

The flow for adding a new country will be like this:
 - add an object to the array, as usual
 - find the corresponding emoji flag, for example at https://emojipedia.org/search/?q=switzerland
 - copy and paste the emoji into the object's `f` property


Thanks for your contribution!

jamietyger:
Hi @ivan-aksamentov 

Thanks for the useful feedback 😄 

image

ivan-aksamentov:
@jamesscottbrown thanks!

All green, LGTM.

@emmahodcroft it's ready for reviewing and merging.

emmahodcroft:
@jamesscottbrown This is super awesome!! 🙌  Thank you so much! I really, really love the way this looks.

emmahodcroft:
@jamesscottbrown & @ivan-aksamentov
Unfortunately one small problem discovered (I previously was working off Ivan's screenshots he sent me) - it seems like Windows 10 may not actually be able to render flag emojis 😞 . This seems to be a long-standing issue, and would mean that anyone on Windows only sees letters instead:
image

I love the simplicity of the emoji Idea, but I'm wondering if we should switch to small icons instead, to ensure they show up across all platforms? ☹️ Alternatively if there's a way of detecting OS we could just use the picture-icons if Windows is detected ( don't know if that's easier or not).

emmahodcroft:
For example, while it would be some work, I could go through and download the [flag image files](https://openmoji.org/library/#group=flags) from OpenMoji: https://emojipedia.org/openmoji/. Or maybe it's possible to get them in a bulk download... I haven't looked too closely yet at options.

ivan-aksamentov:
Oh no! Windows... Is that's in Edge?

We could use SVG icons then. For example, we use this package in Nextclade:
https://www.npmjs.com/package/flag-icon-css

Here is ow we import React components:
https://github.com/nextstrain/nextclade/blob/f1de3c373f8b71cdc20b30e7785d57e5d8dae19b/packages/web/src/i18n/i18n.ts#L16-L25

And then we just use it like this:
https://github.com/nextstrain/nextclade/blob/f1de3c373f8b71cdc20b30e7785d57e5d8dae19b/packages/web/src/components/Layout/LanguageSwitcher.tsx#L56

And that's the CSS:
https://github.com/nextstrain/nextclade/blob/f1de3c373f8b71cdc20b30e7785d57e5d8dae19b/packages/web/src/styles/components/LanguageSwitcher.scss#L3-L19

emmahodcroft:
Yea I know, I'm pretty disappointed with Windows. Apparently Emojis are rendered by OS or something --- so these don't show up for me even in Chrome!

What you've posted above seems like it could be a good solution, though the implementation is a little beyond me. It looks like the image files use the country code for the name though - we have a list of countries already, and I'm sure I could add to this the country codes (to go into `web/data/countryStyles.json`) if that's what we need to figure out which icons to use. I guess once you implement it, it has flags for all countries available, it's just a matter of knowing how to call what you need?

ivan-aksamentov:
There are also packages to resolve country codes from country names, in both python and javascript. So it we might not need to search country codes manually.

I'll let @jamietyger to experiment :)

PR #136, opened 2021-03-30
First Attempt at USA 'Per Country' Page

emmahodcroft:
There is likely a lot to get upset about here in terms of programming, but I couldn't resist trying to put together a page that shows the per-country graphs but for... US states!

This is a work-in progress.

Things Emma can do, but needs to do:
- [x] Actually integrate `scripts/try_state_data.py` into the main running script, or at least one that more formally branches from the main script
- [x] Create a `PerStateIntro.md` and have this go into the page header instead of `PerCountryIntro.md`

Things Emma doesn't know how to do:
- [x] How do we give colours to Variants that aren't normally plotted (on main page), but which I've included here? 
image
- [ ] How to best add colours to the dots next to states (Currently grey) without messing up country colors?

Things to think about:
- [x] How to eventually surface this page? I'd like to have the option of adding something similar for Switzerland, but the header is going to get too crowded. Should any additional 'breakdowns' just be links at the top of the 'per-country' page? (I'm not imagining maintaining a ton of these, and they really have to be interpreted with caution as a lot of the numbers get _very_ small.)

@ivan-aksamentov I'm sorry in advance for the fact that I'm sure this could be done much more neatly! I mostly copied and adapted. Please do tidy what should be tidier! 😬


ivan-aksamentov:
I removed duplicated components and added "Region switcher" component with buttons



Merged US data into countries JSON. They now contain "regions" array as top level, and "World", "United States" and "Switzerland" being regions. New regions can be added here:

https://github.com/hodcroftlab/covariants/blob/c7635f53eca65eba68ab45656d71bb6ed64b5775/scripts/convert_to_web_app_json.py#L342-L346

emmahodcroft:
The last issue remaining here is to make the grey dots for states have colors. However, I don't think this need hold up the merge. I'm hoping we might still get the country flags worked in as pictures by modifications to PR #135 - and if that comes through, then we can do the same here!

PR #144, opened 2021-04-23
Fix typo in 20A.EU1.md

henryaj:
Believe this should read 'N-terminal domain' (pertaining to the 'N-terminus') instead of 'non-terminal domain'.


emmahodcroft:
Thank you @henryaj !

PR #145, opened 2021-04-26
Improve "Defining mutations" section

ivan-aksamentov:
This moves "Defining mutations" section to the sidebar


emmahodcroft:
I really like this @ivan-aksamentov ! Thank you for this really neat idea. I'd like to merge this in, in the next couple days, if you feel its ready? (May do it at the same time as a new variant being added to the site.)

ivan-aksamentov:
@emmahodcroft Yes, it seems to be ready. You might also want to uncomment the additional mutations in `clusters.py`, that were hidden previously.

Issue #137, opened 2021-03-31
Legend says 'week' when it's a 2-week period


corneliusroemer:
The tooltip legend says 'week' when in fact what's displayed is data from a 2-week period.
![image](https://user-images.githubusercontent.com/25161793/113144397-7a6bc980-922d-11eb-9f6c-5f5b62d39cfa.png)

Also, since only one date is provided, I'm not sure whether the 2-week period begins or ends on the date being shown.

Maybe it'd be better to show start and end date together? Or say 2w-period beginning on x, or at least mention it in the description of the site at the top? I couldn't find any explanation thus far.

Issue #109, opened 2021-02-24
variable time-intervals in "per country" plot


TYMichaelsen:
Hi,

When looking at the Danish frequency plot of variants here https://covariants.org/per-country I noticed that the intervals varies in size? so for instance 2020-12-21 to 2020-12-28 is one week and 2020-12-07 to 2020-12-21 is two weeks? I hope it's a mistake! :)

On that note, could the dates be more informative? So for example indicate which ISO-standard week they cover instead of showing a specific date?

Best,
Thomas

emmahodcroft:
Hi @TYMichaelsen Thanks for spotting this. From checking the rest of the data, it seems this one date period (2020-12-21 to 2020-12-28) is not following the pattern we intend! I think this might be an edge-case about how we divide up into the two-week periods. I'll investigate this and see if we can get it fixed!

Regarding the dates/ISO week - I decided to go with the 'actual dates' as for people who don't work with ISO weeks regularly they're quite confusing and not intuitive, but I agree that for those who are familiar with ISO weeks this might make more sense. A good option might be to have a toggle-switch at the top of the page so that users can switch between the two and use whichever is more useful for them! 

Would you mind making an issue for that, and we can see if we can get that implemented?

emmahodcroft:
Hi @TYMichaelsen - I tracked this down to the change in year from 2020 to 2021. Here's how this is done in `datetime` converting from ISO week to date:

```
(2020, 50)  (2020, 12, 7, 0, 0)
(2020, 52)  (2020, 12, 21, 0, 0)
(2021,  0)  (2020, 12, 28, 0, 0)
(2021,  2)  (2021, 1, 11, 0, 0)
```

The issue seems to be that essentially there has to be a little fiddling with ISO weeks, because the ISO weeks don't necessarily divide so evenly into the year. (I don't work with ISO week dates myself so this is all a bit new to me.)

As far as I could tell from researching, there doesn't seem to be an easy fix for this for 2-week increments. I think the best solution might be to just accept that this one point will be 'off' (but it should not repeat until perhaps New Year 2022), and work to enable a toggle to switch between dates & ISO weeks so that this is easier for users to investigate (and hopefully, if they are a bit familiar with ISO weeks, to see the cause in the end-of-year!). 

However, if you're more familiar with ISO weeks and know of ways this is handled by others, I'd certainly appreciate any thoughts you might have!

corneliusroemer:
[...]

PR #148, opened 2021-04-29
Fixes and improvements for date intervals and date formatting

### LINKED ISSUE #147: More regular date lines on Per Country page
Currently the date lines drawn on Per Country are a bit arbitrary - I think that's maybe ok, but there is a big gap between ~October 2020 and the last data point (currently 5 April 2021):
image

It seems like one more line might fit in between these two - but perhaps by the current spacing guidelines that's not the case? It just looks a bit odd to have no other line in between these two time points that are now about 6 months apart.

ivan-aksamentov:
Resolves #147 #137

This PR:

1. Adjusts number of x-axis ticks responsively (depending on screen size) and improves formatting of dates in the ticks, such as it is more human-friendly.

   Some adjustment might be needed until it looks good, for different screens. This can be done by tweaking the desired minimum tick width: https://github.com/hodcroftlab/covariants/blob/438b2fa36a8d7de9f281edab236168c1a20444e7/web/src/theme.ts#L119

   Examples:

    Screen 1920px wide:
	  <a href="https://user-images.githubusercontent.com/9403403/116506280-41228a00-a8bd-11eb-9e84-eb55d2ff4066.png">
	     
	  </a>

    Screen 1280px wide:
	  <a href="https://user-images.githubusercontent.com/9403403/116508129-f7d43980-a8c0-11eb-8929-052fb42a0255.png">
	     
	  </a>

2. Improves formatting of dates in tooltips

	Date is displayed as an interval and in human-friendly form.

	~This implementation assumes that the `week` field in the data points is the **end** date for each 2-week interval.~ UPDATE: The assumption was incorrect and intervals are now changed to begin with a data point and to end 2 weeks after that.

	  <a href="https://user-images.githubusercontent.com/9403403/116507702-1980f100-a8c0-11eb-8f00-c015bbff1409.png">
	     
	  </a>


emmahodcroft:
Thanks Ivan, this is looking good! Particularly the dynamic date interval adjustment 👍 

Unfortunately however, the `week` given in the raw data is the _start_ of the 2 week period. Could we adjust so that it reflects that?

ivan-aksamentov:
Done!

emmahodcroft:
This looks great @ivan-aksamentov  - thank you! There is just one small thing that I suppose we should discuss how important it is. There's an 'issue' about the two week intervals that I use for the per-country plots. You see - technically everything's calculated by 'ISO week' - then I just extend that to being 2 weeks, with a bit of a hack. 

However, what this means sometimes a 2 week period doesn't fit into the slightly hazy way that ISO weeks handle switching from one year to the next - so we end up with periods shorter than 2 weeks when they get converted back to 'date':
https://github.com/hodcroftlab/covariants/issues/109#issuecomment-785836969

>```
>(2020, 50)  (2020, 12, 7, 0, 0)
>(2020, 52)  (2020, 12, 21, 0, 0)
>(2021,  0)  (2020, 12, 28, 0, 0)
>(2021,  2)  (2021, 1, 11, 0, 0)
>```

At the moment this is 'odd' but not hugely problematic as it just means the eagle-eyed spot that one of the time periods is just 1 week rather than two.

However, with the labels, it's assumed these are always a clean 2 weeks - when for one week in December, it's not!:
image
image


This should only ever happen around the New Year (aka this shouldn't be a problem again until Dec 2021). What do you think?

PR #152, opened 2021-05-02
Add 617.1 and 617.2 lineages as 20A/S:154K & 20A/S:478K

emmahodcroft:
Adds variant tabs & entries & plots for 154K & 478K, as well as a Nextstrain build (currently 1 build for both, both link to that build). Can separate out in future.


lhelseth:
Emma, When previewing this PR I noticed the bar graph of countries at https://cov-lineages.org/lineages/lineage_B.1.617.html I noticed an apparent typo in the legend; the label for the USA reads "TED States of America".  It's that way on other variant pages as well.  Mousing over the graph shows the full "United States Of America".  Perhaps shorten to  "United States" or "USA" or expand the height of the legend?  Thanks.  Go Frogs!

emmahodcroft:
Hi @lhelseth - thanks for reaching out - and yes, Go Frogs! 
Good spotting - I'm guessing this is getting cut off because of the length of the label. However, cov-lineages.org isn't my site - it's maintained by the Pango group - I just link for those interested in seeing some different information. I don't know if they have a github or a contact page, but it would be good to reach out to them with what you spotted!

PR #153, opened 2021-05-02
Update 20A.S.478K.md

jcbarret:
I think this is a typo: T19Q should be T19R?


emmahodcroft:
Thank you @jcbarret - you are right! Good spotting, and thanks again!

emmahodcroft:
This should be live now - thanks again!

PR #159, opened 2021-05-11
Try 1 Week Intervals for Per Country

emmahodcroft:
There is a long-standing issue where using 2-week periods (as currently used) for Per Country can be problematic at the end of one year and beginning of the next (#109).

This came up again with PR #148. One possible fix is just to switch over to one-week intervals instead of 2-week intervals. This PR gives this a try.

See the deployment for all graphs, but for a few examples...
2 weeks:
image

1 week:
image


emmahodcroft:
Decided this is too jerky. Instead, we've improved the date plotting and shown the interval covered by each point. Also, Emma will look at a different way to count weeks so that it better covers 'year-changes'.

PR #165, opened 2021-05-25
Switch to using `mutation_summary` file

emmahodcroft:
Currently CoVariants.org uses `sequence-diagnostics_gisaid.tsv` - as this is being phased out of the `ncov` workflow, this PR switches to using `mutation_summary_gisaid.tsv` which contains much of the same information (though slightly differently represented in some cases).

In future, using this file will allow specification of the exact mutation one wants to select, rather than just 'any mutation at this nucleotide position', and also in theory to use AA mutations. However, importantly, if there's some kind of frameshift that impacts translation, then AA mutations may not show up, whereas nucleotide mutations always appear.

This is why I'm sticking to using nucleotide mutations for the foreseeable future.

As part of this PR - since `mutation_summary` is 1-based, whereas `sequence-diagnostics` was 0-based, all the SNP definitions in `clusters.py` have been changed to 1-based.

### DIFF
--- a/scripts/allClusterDynamics_faster.py
+++ b/scripts/allClusterDynamics_faster.py
@@ -56,6 +56,7 @@ from clusters import *
 from bad_sequences import *
 from approx_first_dates import *
 import os
+import re
 
 
 def get_division_summary(cluster_meta, chosen_country):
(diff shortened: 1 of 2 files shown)


emmahodcroft:
I've tested this on all of the clusters now, and the files are not modified except for `ORF1a.S3675`, `S.H69-` and `S.Y144-` - these are all builds based upon gaps (deletions). In each, the number of sequences differing is usually <10, in cases where countries have a large number of sequences is at most <100. 

Upon closer inspection, it appears that there are likely 2 causes here. The `sequence-diagnostics` gives gaps as a range, but it's unclear whether this includes or excludes the starting & ending position given - so there's ambiguity here which means that some sequences may not be being captured. When I expand this range to positions, it's also possible I'm not expanding that range to match what's intended by the listing in the file (inclusive/exclusive of last position, in particular).

However, this is resolved by using `mutation_summary` because gaps are given explicitly, and given the small number of sequence impacted, I'm happy to proceed and assume the few sequences affected were because of confusion about what was in a gap 'range'.

PR #168, opened 2021-06-02
add new WHO Labels to variant table

jamesscottbrown:
This adds the new [Greek letter variant names](https://www.who.int/en/activities/tracking-SARS-CoV-2-variants/) to the table of variants:

who-labels

### DIFF
--- a/scripts/name_table.py
+++ b/scripts/name_table.py
@@ -4,6 +4,7 @@ import urllib.error
 name_table = [
     {
         "clade": "20E (EU1)",
+        "who": None,
         "lineages": [
             {"name": "B.1.177", "url": None},
         ],
@@ -13,6 +14,7 @@ name_table = [
     },
     {
         "clade": "20A.EU2",
+        "who": None,
         "lineages": [
             {"name": "B.1.160", "url": None}
         ],
@@ -20,6 +22,7 @@ name_table = [
     },
     {
         "clade": "20I/501Y.V1",
+        "who": "Alpha",
         "lineages": [
             {"name": "B.1.1.7", "url": None}
         ],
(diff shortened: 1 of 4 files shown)


emmahodcroft:
@jamesscottbrown you are an angel and a mind-reader! I had this on my list to make an issue for, and here you have come and presented a PR already! 😁 

@ivan-aksamentov Is there a way we can get this to deploy, so we can check it out, or is that not possible for 'outside of repo' PRs?

emmahodcroft:
Yea, I thought this was weird as I definitely remember looking at deploys from other people's PRs. But I'm never sure if I've missed some critical change in terms or something!

Hopefully this is related to the Github issues, and will resolve automatically. If not when it seems like issues have been resolved, maybe you could push some 'nothing' commit to get it to try again. But we can wait a bit and see if we get a deployment!

ivan-aksamentov:
Not sure what happened, but looks like it works now. I did nothing.

Regarding the code, looks good to me, thanks @jamesscottbrown. I haven't checked the exactitude of the science part though, that's on @emmahodcroft.

If there's not always one-to-one mapping between 'covariants variants' and 'WHO variants', then the entry might need to be an array, similarly to lineages. But we can worry about it when it happens.

In going further, the information from this table can be merged into `clusters.py`, in order to show the name aliases on other pages, tooltips, variants's sidebar etc, as discussed with Emma previously.

emmahodcroft:
Sorry for the delay on this but I think it's looking great!

@ivan-aksamentov You raise some good points. As you might have seen, we're about to revamp the Nextstrain clade system, so I think it's best not to make any changes now, and I'll have a think about the best way to incorporate this into CoVariants - then we can see about incorporating more deeply into `clusters.py`!

(I'm afraid I will soon be sending you back to the drawing board on the Variant Badges for new color patterns once again! 😬 😁 )

PR #169, opened 2021-06-11
Change and reorder Clade Names!

emmahodcroft:
This updates Clade Names to match those used on Nextstrain. Will require a few steps of changes before all is probably working as it should be - but having deploys will help.


emmahodcroft:
Of course... this doesn't have the circleCI config. I can dig into this.

PR #171, opened 2021-06-17
Use monospace font for variants, prettify name table

ivan-aksamentov:
Switched the font for variant names in some places to monospace. This way the sections in new variant names are nicely aligned.

Current font is "Source Code Pro"

We might try other, prettier fonts too. I would think that a monospace sans serif font would do the best here. Serif fonts jump out immediately, because the rest of the page is sans. Here are some random font viewers:

https://fonts.google.com/?category=Sans+Serif,Monospace
https://www.1001fonts.com/monospaced-fonts.html
https://www.fontsquirrel.com/fonts/list/classification/monospaced
https://fontsarena.com/monospace/
https://en.wikipedia.org/wiki/List_of_monospaced_typefaces

Screenshots:
---

---

---

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -111,8 +111,8 @@
     "reselect": "4.0.0",
     "serialize-javascript": "5.0.1",
     "styled-components": "5.1.1",
-    "typeface-open-sans": "1.1.13",
-    "typeface-source-code-pro": "1.1.13",
+    "typeface-droid-sans-mono": "0.0.44",
+    "typeface-open-sans": "0.0.75",
     "url-join": "4.0.1"
   },
   "devDependencies": {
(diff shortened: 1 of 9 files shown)


emmahodcroft:
Ok, I had a good look around and here are my two favourites. I found that I don't really like either having 'dots' or 'lines' through zero (ex: `0`) in the table - it's displeasing to my eye in the more-casual/less-program-y nature of the table.

So my top pick would be:
- [Droid Sans Mono](https://www.1001fonts.com/droid-sans-mono-font.html)
image

With second place:
- [NK57 Monospace](https://fontsarena.com/nk57-monospace-by-typodermic-fonts-inc/)
image

But I like the first one better as it looks cleaner and more 'normal' to me (which I think is what we're going for here, since it's not actually programming or numbers).

ivan-aksamentov:
>I don't really like either having 'dots' or 'lines' through zero [...] in the more-casual/less-program-y nature of the table

Agreed! I was also looking for one of those.

Will switch to `typeface-droid-sans-mono` package now.

PR #173, opened 2021-06-18
Make cluster button sidebar width fixed

ivan-aksamentov:
This makes button sidebar width fixed, such that it is easier to find an appropriate width for all screen sizes. The layout of variants and home pages had to be revamped for that.

Now only the central text content section is responsible. Both sidebars are of fixed width. 

For smaller screens we tweak the button width slightly, along with height and font size.


emmahodcroft:
From playing around at different screen sizes this PR looks good to me, if you feel it's ready to be merged into master!

PR #179, opened 2021-06-21
Use consistent cluster colors

ivan-aksamentov:
Attempts to address: https://twitter.com/hnoejgaard/status/1406896874766929922

The challenge is that the AreaPlot area segments are transparent (default opacity is 0.6, configurable). If we make the sidebar color boxes transparent it looks fins and the colors match. However, in the tooltip, the table is striped, so transparent colors get tinted by the row colors.

We need to ~find a way to~ produce a non-transparent version of the color.
**UPDATE**: figured it, in

https://github.com/hodcroftlab/covariants/blob/fc704b60edcf8121059f974bcdc4fb92e268dce8/web/src/io/getClusters.ts#L76-L95

**UPDATE2**: Nope, the colors are still slightly different.

One disadvantage of shifting to these "transparent" versions is that we are showing the color that is different from what comes from the data source. So it's confusing for maintainers, because they don't see the colors they put into the actual data.

Another way of tackling this problem is to use original, opaque colors everywhere, including the plots. It looks a bit angry though, due to the bright red colors:

And to fix that, we could make the original colors more pastel. I think this would be the cleanest solution.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -61,12 +61,14 @@
     "stills": "yarn run:node tools/generateStillImages.ts"
   },
   "dependencies": {
+    "@types/color": "3.0.1",
     "animate.css": "4.1.0",
     "autoprefixer": "9.8.6",
     "axios": "0.21.1",
     "bootstrap": "4.5.2",
     "bootstrap-icons": "1.2.2",
     "classnames": "2.2.6",
+    "color": "3.1.3",
     "connected-next-router": "3.1.0",
     "core-js": "3.6.5",
     "css.escape": "1.5.1",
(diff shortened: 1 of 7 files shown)


emmahodcroft:
I think the last PR maybe messed up how the colors looked for me? Now they look like this - not matching (legend is too light now): 
image

But if I look at the deploy previous to that (3c785e6) then they seem matching:
image


Alternatively if this is all a bit much, I could set all the colors to the more pastel version in `clusters.py`, and we plot the graphs at 100% opaque - I can probably just dropper-grab the colors from the 1st deploy :) - let me know!

ivan-aksamentov:
These are both different colors! I have a problem with blending colors in both cases.

Yes, I could just run the script, that produced the first results on all colors and replace the colors in clusters.py (or wherever they happen to be nowadays)

ivan-aksamentov:
Superseded by #186

PR #180, opened 2021-06-21
fix: safari cluster buttons misaligned

ivan-aksamentov:
Splitting the `flex-basis` out of the `flex` statement seems to be working correctly on Safari 14, while the compound variant does not.

Hello, Apple, anybody home?

Before:
before

After:
after

iPhone 7 iOS 12 looks fine:

### DIFF
--- a/web/.browserslistrc
+++ b/web/.browserslistrc
@@ -3,6 +3,7 @@ defaults
 cover 99.5%
 since 2013
 ie >= 10
+safari >= 10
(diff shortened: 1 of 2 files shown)


emmahodcroft:
Thank you Ivan! I'm happy to merge this in, if you're happy with it!

ivan-aksamentov:
Yes, it should be ready

PR #181, opened 2021-06-22
improve speed of some of the steps

emmahodcroft:
Speed and logic improvements to the code

- Removes 'bad sequences' more efficiently at start
- Removes duplication of trying to remove sequences with invalid dates & bad sequences twice
- Filter mutations file to match metadata, so if a sequence has been removed from meta, it is also out of the mutations file
- (most significant) Reduces time to search for matching sequences in the mutations file
- (Unrelated) Clears some warnings associated with setting to copies of slices of dataframes


emmahodcroft:
I've done a full run with this and all resulting files (barring some binary ones like JPG) are identical.

PR #183, opened 2021-06-23
More Speed Ups

emmahodcroft:
Reworks the section which bins data by date to be much much much faster (on all clusters, reduces from 30 min to 5 min).

Also tidies up screen output somewhat (removes some old things), and adds some timing so that can keep an eye on what bits take how long.


emmahodcroft:
This has been tested on a full dataset.

PR #186, opened 2021-06-24
Pastel-ize Colours to Improve Plotting Matching

emmahodcroft:
Per Country plots are plotted at partial opacity to make them look less 'angry', but this means the colors & legends don't actually match, which can cause confusion.

#179 attempted to address this by changing the colors of the legend, but this turns out to be quite a challenge. 

As an alternative approach, I've changed the colors of the clusters to be the 'pastel versions' used in plotting. We can now try plotting the graphs at full opacity, which should mean they look the same as now, but the legends will always match, too.

Can you add the full-opacity plotting to this @ivan-aksamentov and we can see how it looks?


emmahodcroft:
Converted to a draft as shouldn't be merged until we can see how it looks with the opaque plotting.

ivan-aksamentov:
>Can you add the full-opacity plotting to this @ivan-aksamentov and we can see how it looks?

Done! Red colors still look a bit irritated. Perhaps we calm them down a bit more?

emmahodcroft:
Actually I think this is looking very good! The colors in the charts match what we have now (with the darker colors but then making them transparent-ish), so I'm happy with it.

emmahodcroft:
I may have to do some more extreme color modifications in the future, as we get to more variants (already stretching things a bit), so I think for the moment I'll leave as-is, and can think about colors more seriously in the future, if changes need to be made... :)

PR #189, opened 2021-06-28
Sort country and cluster web data by week

ivan-aksamentov:
This is to avoid time going backwards in some of the plots.


emmahodcroft:
I've tested this locally and I can't reproduce any of the issues from this morning. I think this squashes the 🐛 !

PR #190, opened 2021-06-28
Render plot cursor above area segments on country plots

### LINKED ISSUE #188: Plot mouse-over-line on Per Country Page on top of Graphs
In an attempt to get colors of legends & plots to match better, we switched to using more pastel colors throughout CoVariants and then plotting at full opacity (see PR #186 ).

This works great, but I just noticed that it means the 'mouse-over-line' (and dots) are no longer easily visible, as they are behind the now-opaque graph fill. 

Before the change in opaqueness/colour:
image

After:
image
(No line, also dot partially obscured)

It's possible to see the older behaviour in deployed PRs from prior to #186 merge, like [here](https://covariants-k4u70foev-hodcroftlab.vercel.app/per-country).

Possible solution:
Could the line and dots be plotted on top of the graph fill?

ivan-aksamentov:
Resolves #188

After #186 the cursor line became invisible because area segments on the plot are opaque:

image

This PR renders the `<Tooltip/>` components (which includes the cursor line) above area segments.

There is a new problem though. The "active circles" (the circles showing the value on the intersection of cursor line and area segment) are now partially obscured by area segments:

image2

This is due to them being rendered intertwined along with area segments, so that the previous circle is obscured by the next segment:

(note `area` and `active-dot` element ordering)

 I am not sure we will be able to fix that easily.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -102,7 +102,7 @@
     "react-super-responsive-table": "5.2.0",
     "react-toggle": "4.1.1",
     "reactstrap": "8.5.1",
-    "recharts": "2.0.0-beta.8",
+    "recharts": "2.0.9",
     "redux": "4.0.5",
     "redux-logger": "3.0.6",
     "redux-saga": "1.1.3",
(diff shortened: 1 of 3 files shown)


emmahodcroft:
Thank you @ivan-aksamentov ! The plan now is to merge this in, but continue to think about the problem more generally, including considering some ideas from #179 .

PR #191, opened 2021-06-29
Fix crash in old safari on missing ResizeObserver

ivan-aksamentov:
Plot rendering depends on ResizeObserver. Older Safari don't have ResizeObserver support and crash when navigating to one of the pages containing plots.

This PR adds a global polyfill (`resize-observer-polyfill` package)  so that Safari display plot pages correctly.

Just tested this PR on Browserstack. Looks like it now works on things as far back as iPhone 6, iOS 11, circa 2014. A blast from the past!

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -111,6 +111,7 @@
     "reflect-metadata": "0.1.13",
     "regenerator-runtime": "0.13.7",
     "reselect": "4.0.0",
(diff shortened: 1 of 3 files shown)


emmahodcroft:
Thanks Ivan! Have also tested on Browserstack, and seems to work well!

PR #209, opened 2021-08-29
Add US state code "flags"

richardgoater:
Hi folks! Hope you're well. This is one idea I mentioned a while back.

I think the types for the flag components are slightly unresolved atm, it probably needs a generic flag interface.

Let me know what you think 😃


ivan-aksamentov:
Thanks, Richard @richardgoater 

Looks pretty!

On the technical side though, should we just hardcode all the states? We don't expect any more states added, right?

Hardcoding would remove the 3rdpary dependency (and the really old version of lodash it depends on for some reason) and should simplify the code significantly. I mostly worry about all the null checks.

Here is what I would do:
 - take the current code of the PR
 - console.log all the pairs
 - copy them from console
 - paste into the text editor and make an object from them
 - make a function to lookup the object
 - call the function to get the letters

Or do you plan on using the library for something else?

richardgoater:
Hi @ivan-aksamentov, I completely agree, I was just being lazy. Thanks very much for writing up a "migration guide"! I should be able to do it tomorrow.

richardgoater:
think that should be done now, let me know what you think 👍🏽

emmahodcroft:
If we can resolve the couple of merge conflicts on this, I'm happy to merge it in!

richardgoater:
Thanks! I'll take a quick look :)

emmahodcroft:
Looks like Ivan's on it ;) I'll wait for the deploy after the merge, but if all seems to have gone without a hitch, we can merge it in!

ivan-aksamentov:
Yeah it conflicted with my continent flags feature really badly. I think I might have resolved it, but ple

I also made us state codes to be in monospace font (all letters have same width) so that narrow letters don't make things wobble as you go through the list. (e.g everything that has and "i" or "j", like Indiana or New Jersey). How do you like it? 

<table>
<tr>
<th>
Before
</th>
<th>
After
</th>
</tr>

<tr>
<td>

<td>

</td>
</tr>

<table>

Also we don't have a code for the Mariana state. Never heard of it and don't know what should be there. 

Please double check everything, because merge was somewhat painful and required a few attempts.

ivan-aksamentov:
Alright, according to 
https://en.wikipedia.org/wiki/Northern_Mariana_Islands
"Northern Mariana Islands" should have code "MP". I added it. Sounds good?

emmahodcroft:
I've gone through this and it looks great to me Ivan - thanks! Not sure I'd have noticed the font spacing, but I agree it looks nice with it all taking up the same amount of space! And I've never heard of the Mariana Islands either... 🤔 

Shall we wait for the deploy to finish to ensure all is good? Then we can merge.

emmahodcroft:
Ok, deploy went fine - I'll merge in!

emmahodcroft (approved):

PR #210, opened 2021-08-30
Demo of ideas for polishing

richardgoater:
Really sorry about the monster PR - at this stage it's just a demo. Would be happy to deconstruct it if you'd like to keep any of the ideas, and I might still add a few more but wanted you to take a look :)


richardgoater:
Hi folks, here's the write-up as promised:

* This is the "fewer shadows" option that I mentioned (the grey background did not work out!). I think it allows shadows to draw attention to important items and the outline of the page is deferred to whitespace.

* Text has been constrained to 100 characters wide. This is to make sure lines are not too long on wide screens, which should keep the content easy to read.

* The FAQ page has been polished the most. The rules for it should apply to all editable sections, fingers-crossed they are general enough.

* Page headings have been standardised and given more room. Padding and margin has been cleaned in key places.

* I have attempted to improve the margin around the charts to let them fit the card and make space for the tick at the top of the Y axis. X axis ticks also have a bit more space.

* I have also tried to fix the domain to match the ticks. This is a cleaner and more reliable solution, but as we can see starting from April 1st 2020 doesn't fit perfectly so this may need more work.

Comments welcome as always!

emmahodcroft:
So I've had a brief look into the code output, but I can't find in any of the output files that there are any dates before "2020-04-27".

So, I'm left unsure how we're getting signs that data exists before this cutoff...? Can anyone give a tip for where the 'earlier date' is coming from (what file) and what it is?

ivan-aksamentov:
We'll need to check what's happening in the app then. Maybe I did something unexpected knowing that `params.json` sets the correct domain from `params.min_date`.

richardgoater:
I think I see the issue in `getParams.ts`, it is rounding the ticks to the start of the month, but it should start at the month after the initial date. We should then use the specified domain as before. You were falsely accused @emmahodcroft, I do apologise!

emmahodcroft:
No worries, I can completely believe that I might not have been being firm enough in the data cut off, so it was good to check and ensure nothing suspect was happening!

richardgoater:
I only realise now what the code was really trying to do, I started on the wrong path.

Just noticed that this will also set the ticks on the Per Variant page to start at May 2020. We could do the adjustment only for the Per Country page if that would better. We could also consider making this configurable if e.g. the first tick being June would look better.

ivan-aksamentov:
Richard fixed the problem with a gap in the beginning of the plot. I updated this PR for the latest changes on master. I think it is ready to be merged. @emmahodcroft Please check last one time.

ivan-aksamentov (approved):

emmahodcroft:
Apart from the merge conflict, I've checked through and it looks good to me!

emmahodcroft (approved):

ivan-aksamentov:
Yep, this is what happens when PRs are hanging for too long.
In this case just a little import conflict. I am merging.

emmahodcroft:
Thank you @ivan-aksamentov ! I'll wait until this builds on master and then push everything live. I'll do a little tweet to tag Richard & thank him for his efforts!

PR #211, opened 2021-08-30
Make tooltip settings sticky

richardgoater:
Just a little idea to keep the tooltip settings in view when scrolling through the charts :)

The removal of the overflow is a risky one though, hopefully it doesn't break anything 🤞🏽


emmahodcroft:
Going to go ahead and merge this in, as discussed yesterday - please do let me know if we see anything that might be a consequences of getting rid of the overflow!

PR #219, opened 2021-10-21
feat: show mutation counts

ivan-aksamentov:
We want to display counts and frequencies for all mutations in every cluster.

This adds some fake data and the initial sketch of the tabular representation for this data.

<a href="https://user-images.githubusercontent.com/9403403/138282827-20f14d93-2ca2-4497-bd79-d2254dda7e6d.png">
	
</a>


emmahodcroft:
This is very exciting! We probably do want the table to sit a little lower on the page (below the text, I guess), but I guess it's likely modular so can be easily moved. 
I think we will need to come up with a way to remove the defining mutations from the table - otherwise they're taking up too much of the table. I can see if I can do this on my end while generating the data. 
Might it be worth seeing if we can have the 'S' table and the 'other gene' table side by side? Might make it a little more space-efficient horizontally. But I guess it would have to be considered what happens on mobile.

ivan-aksamentov:
> I guess it's likely modular so can be easily moved.

Yes. As easy as moving this block up and down:

https://github.com/hodcroftlab/covariants/blob/4698625a36160176268a10f820c2d9d038aaad1e/web/src/components/Variants/VariantsPage.tsx#L103-L107

I just wanted to avoid scrolling while we iterating on how it looks.

We could probably somewhere in between the content and the proteins?

https://github.com/hodcroftlab/covariants/blob/4698625a36160176268a10f820c2d9d038aaad1e/web/src/components/Variants/VariantsPage.tsx#L125-L135

Or below the proteins? Feel free to play locally to see how it looks.

It'a s bit space-inefficient though, as there's a lot of blanks to the sides of the table.

> can have the 'S' table and the 'other gene' table side by side?

Yes, I'll make it happen.

> remove the defining mutations from the table

We definitely need  to do this offline. The app is already too sluggish. Performing additional computations on every page load for every user is unnecessary because the data is static.

I am currently also partitioning mutations to S and non-S mutations in JS:
https://github.com/hodcroftlab/covariants/blob/4698625a36160176268a10f820c2d9d038aaad1e/web/src/components/MutationCounts/MutationCountsSummary.tsx#L52
which is also unnecessary, because it's static data. We could also do this offline, producing 2 arrays, one for each table.

The goal is to have the final 2 arrays which can be displayed as is.

emmahodcroft:
Just FYI, @MoiraZuber will start taking a look at this next :) 

(This message mostly just to keep people up-to-date on what she's up to!)

emmahodcroft:
@MoiraZuber You can head to [here](https://github.com/hodcroftlab/covariants/pull/282) for the new version of the PR :)

ivan-aksamentov:
This PR's diff got a bit out of had, I resubmitted it with @MoiraZuber's permission as: 
https://github.com/hodcroftlab/covariants/pull/282

The branch is the same. Let's continue the work there.

PR #221, opened 2021-11-11
Add region (continent) filters

ivan-aksamentov:
Adds continent checkboxes to the filtering sidebar, such that countries can be toggled  en masse by continent.

The two-directional updates between country and continent checkboxes required changing the underlying state structure. So I moved a lot of things around, And while on it I decided to improve the code  of everything that touches this state. So now the codebase should hopefully a little bit better, at the expense of non-reviewable huge diff in this PR.

Example: everything is deselected and then Europe is selected:


emmahodcroft (approved):

emmahodcroft:
Thank you Ivan, this is very cool!

PR #222, opened 2021-11-16
fix: add missing flags

ivan-aksamentov:
This adds a couple of country codes to the list of replacements, so that the flags are looked up correctly.

The incorrect lookup is due to us using country names which are not compliant with the ISO standard.


emmahodcroft:
This looks great to me Ivan, I'm happy to merge if you are?

PR #237, opened 2021-12-04
name_table.py, MutationBadge.tsx: repair global reports polling and display logic

jcahill:
Tweak to `update_name_table()` to add a bit more checking for changes in global report page availability.  General notes below.

## covariants homepage error

This report linked in the name table does not exist:

- https://cov-lineages.org/global_report_B.1.617.1.html

## all global reports

As of now, the following pango global report pages exist:

- https://cov-lineages.org/global_report_A.23.1.html
- https://cov-lineages.org/global_report_B.1.1.529.html
- https://cov-lineages.org/global_report_B.1.1.7.html
- https://cov-lineages.org/global_report_B.1.351.html
- https://cov-lineages.org/global_report_B.1.525.html
- https://cov-lineages.org/global_report_B.1.617.2.html
- https://cov-lineages.org/global_report_P.1.html

## future global reports

Global reports for lineages not of VOC status were deprecated in https://github.com/cov-lineages/lineages-website/commit/ddd6db717fcf20a318640c693cbf490c6cd7d0e0.
Two reports excluded by this change have not yet been removed.
These are:

- https://cov-lineages.org/global_report_A.23.1.html
- https://cov-lineages.org/global_report_B.1.525.html

### DIFF
--- a/scripts/name_table.py
+++ b/scripts/name_table.py
@@ -204,9 +204,6 @@ def does_url_respond(url: str):
 def update_name_table(name_table):
     for datum in name_table:
         for lineage in datum["lineages"]:
-            if lineage["url"] is not None:
-                continue
-
             name = lineage["name"]
 
             # Make a request to a hypothetical URL where global report might reside.
--- a/web/src/components/Common/MutationBadge.tsx
+++ b/web/src/components/Common/MutationBadge.tsx
@@ -286,7 +286,7 @@ export interface LineageLinkBadgeProps {
 export function LineageLinkBadge({ name, href, prefix, report }: LineageLinkBadgeProps) {
   const url = useMemo(
     // prettier-ignore
-    () => (href ?? report ? `https://cov-lineages.org/global_report_${name}.html` : ''),
+    () => (href ?? (report ? `https://cov-lineages.org/global_report_${name}.html` : '')),
     [href, report, name],
   )
   const tooltip = useMemo(() => `Pango Lineage ${name}`, [name])


emmahodcroft:
@ivan-aksamentov  - it's not intuitive to me, what the bit of code this PR removes was doing (seems like it shouldn't have been there?) -  but perhaps there's a reason it was added. What do you think?

ivan-aksamentov:
@emmahodcroft Currently:
 - if an entry in the name table has `.lineage.url` set explicitly, then it is used as URL behind the corresponding badge in the table
 - if not, the script attempts to build a URL to cov-lineages.org from lineage name and to make an HTTP request to it. If it responds with "200 Success", then there's probably a page with lineage details, and so this URL is used.
 - if it does not respond with 200, then the URL is `None`

In this PR the check for the manual explicit URL is removed, changing  the logic to:
 - the check is always performed and if the automated URL responds, it is used regardless of whether a manual URL is present (overwrite manual)
 - if it does not respond:
    - if manual URL is set in `.lineage.url`, then it is used
    - otherwise the URL is `None`

Currently one of the manual URLs on master and release is broken (links to a "404 not found").

This is the full code on master:
https://github.com/hodcroftlab/covariants/blob/d60d731974fc6bb69d54be593b879f04a26abb9c/scripts/name_table.py#L204-L218

ivan-aksamentov:
In fact, I just checked and something weird is happening. 

`nameTable.json` contains your manual URLs, as expected. For example for Delta. However, the actual table on home pages is showing something completely different and it leads to a 404 page. This happens on all deployments and locally. I am not sure where these URLs come from. Will investigate. Tell me if you know something about it.

jcahill:
> However, the actual table on home pages is showing something completely different and it leads to a 404 page. This happens on all deployments and locally. I am not sure where these URLs come from.

This is what alerted me to the need to look into how the script is setting URLs. It's this one, as mentioned in OP without elaboration: https://cov-lineages.org/global_report_B.1.617.1.html

Looking again, there's still one layer of weirdness beyond simplifying this check, unless the one bad url was simply the result of running/calling the script twice by accident (i.e. the mutable element is updated on the first run, some entry no longer satisfies `None` condition, second for-loop is filtered in unexpected way by this). However, always checking for a global reports page is an overall improvement to robustness.

jcahill:
I haven't yet heard back, over on the cov-lineages website repo, whether the shift to VOC-only report pages implies that VOI reports will be deleted. The outstanding VOI reports aren't being updated while the VOC ones are, though.

ivan-aksamentov:
@jcahill 
> unless the one bad url was simply the result of running/calling the script twice by accident

But the resulting JSON file containing the expected results:

https://github.com/hodcroftlab/covariants/blob/d60d731974fc6bb69d54be593b879f04a26abb9c/web/data/nameTable.json#L51

That is, Delta contains:
```
https://cov-lineages.org/lineages/lineage_B.1.617.2.html
```

but the website links to

```
https://cov-lineages.org/global_report_B.1.617.2.html
```

So there is weirdness beyond the Python script. It's something in the app as well.

jcahill:
@ivan-aksamentov then it has to be badness in MutationBadge.

2021-12-06_019746_Developer Tools — CoVariants — https___covariants org_

https://github.com/hodcroftlab/covariants/blob/d60d731974fc6bb69d54be593b879f04a26abb9c/web/src/components/Common/MutationBadge.tsx#L286

ivan-aksamentov:
@jcahill Hm, interesting. Do you see why it falls back to that URL is `useMemo()` even if `entry.url` is passed through `href`? I think this `entry.url` should be the thing from JSON.

https://github.com/hodcroftlab/covariants/blob/d60d731974fc6bb69d54be593b879f04a26abb9c/web/src/components/Common/NameTable.tsx#L81

jcahill:
@ivan-aksamentov  It's expression grouping. Fixed in latest commit.

emmahodcroft:
I think this went over my head a few messages ago, but I'm happy to improve the odds we have functional URLS! I don't always have time to search for them, so this can be an improvement.
The only thing that would be good to keep in mind is that sometimes I build locally while without internet - just to ensure there's a reasonable time-out so it doesn't stick for ages, but I don't think this has been a problem for a long time.

jcahill:
@emmahodcroft The change to the python script doesn't substantially change connectivity requirements while building. The change to the badge link styling fixes the spurious global report links in the final web view. If either you or ivan have a minute to review, the [preview](https://covariants-7wjj6qos7-hodcroftlab.vercel.app/) now has fixed links.

ivan-aksamentov:
@jcahill Thanks for investigating this Jeremy!

I merged the Typescript part, but kept the Python logic as it was on master. The idea is that the online URL checks are optional and Emma can override them with manually set values.

I think we should:
 - split the online logic into a separate file + makefile target to make Emma's work more convenient offline. This thing is merely a nice to have and should not be blocking or delaying the workflow.
 - try and see if we should use the new URLs of the form
	  ```
	  https://cov-lineages.org/lineages/lineage_B.1.1.529.html
	  ```
	

P.S. Please feel free to add yourself to the list of contributors as described in #48

ivan-aksamentov:
Another note is that `name_table.py` is easily derivable from the `clusters.py`. Either during offline processing or even in the web app. Having a separate file with basically the same information, but partial, just introduces duplication and increases probability of mistakes (when we update one place but forget to update the other). This file should have never existed. There is a few other files that have the same problem. That needs to be cleaned up eventually.

Issue #30, opened 2021-01-19
Remember country/variant filter state in the URL params


ivan-aksamentov:
We might synchronize the filter checkboxes on the sidebar of plot pages with URL query params.

This would allow:

 - sharing filtered views of these pages
 - bookmarking

which might facilitate sharing, discussion and tracking the information of interest for users.

For example, we might have a `countries` and `variants` params as follows

```
/per-country?countries=France,Germany,Italy&variants=20A.EU1,20A.EU2
```

Navigating to this URL would render the "Per country" page with only France, Germany and Italy and 20A.EU1 and 20A.EU2 variants enabled in filters.

and if user further unchecks 20A.EU1 and Italy and adds Latvia, then the URL in their address bar will become:

```
/per-country?countries=France,Germany,Latvia&variants=20A.EU2
```

emmahodcroft:
I think this would be very cool to have for linking to particular countries in particular, on Per Country page. Or for a state on the USA version of the page - etc.
How hard would this be to implement, is it a really biggie? (I feel like it might be)
I'd be happy to start with just showing the displayed set of countries/regions (not saving URL state for the variants selected).

emmahodcroft:
Coming back to this because just got a request from someone on Twitter.

ivan-aksamentov:
Yes, this will require moving internals around quite a bit. Not immediate. I might try it sometimes.

an-ko-7:
Hey Ivan, I'm the one who spoke with Emma. Since I've got some experience on web dev and time, I could take a shot. I've got it running offline and I've got the basic structure of the project. Any specific tips?

ivan-aksamentov:
Hi @an-ko-7,

How I imagine it:

You'd need to change the `places` state here:
https://github.com/hodcroftlab/covariants/blob/6d7d642ccb43a7f54911b9bf09f95e23a0c1d2ff/web/src/components/CountryDistribution/CountryDistributionPage.tsx#L37
depending on what query params are, and vice versa.

I'd probably use the Next.js native `useRouter()` to get and set URL params. And/or perhaps also a query string library. I think we might already have one in the dependencies.

There is a few things that make things complicated:

 - `continents` and `countries` depend on each other within `places` state, so you need to remember to update both when one have changed. There are `toggleContinent()` and   `toggleCountry()` helper functions which handle checkbox events. For URL params you might need new version of these functions that toggle multiple countries/continents.

[...]

PR #240, opened 2021-12-08
Linking to regions and countries

an-ko-7:
E.g. http://localhost:3000/per-country?region=United_States&countries=Florida~South_Dakota


ivan-aksamentov:
@an-ko-7 It is best practice to write a few words about your PR or at least link to the issue it resolves.  Issue https://github.com/hodcroftlab/covariants/issues/30

The diff should absolutely be fixed. The contribution is not useful if it corrupts production data and if it is hard to review. I don't see why we need to delay that.

If you cannot fix the diff, then feel free to resubmit the PR.

ivan-aksamentov:
From a quick glance over the link to the diff you posted in the issue, here are my thoughts. Perhaps you have already fixed some of these issues since. But I have reviewed only the link you provided.

 - I am hesitant to accept and then maintain the custom logic for the extraction of URL params. As I mentioned in the issue, Next.js has a very well implemented router and it should be preferred. Have you tried it? What went wrong?

 - Why use tildas and replace spaces with underscores? Again, I am hesitant to add and maintain a logic that is unnecessary. I think we should just URL-encode the list. Modern browsers should be able to show a prettified version of it (i.e. the should URL-decode). But need to be checked and Emma need to try it.

 - The idea of the feature was indeed for updates to be bidirectional. I don't expect  average user of covariants to be familiar with URL params, and even if they are willing to write it by hand, they wouldn't know the syntax rules you choose.

 - In the country enabling/disabling logic I think this is doing way to much. Here is what I would do:
    - take params from the router
    - parse and validate URL params as needed, producing a [`Set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) of countries
    - wrap initial state object into a function that sets `enabled` to `Set.has(countryName)` instead of always `enabled: true`
    - use `Set.has()` to decide whether a country need to be enabled

   If you think that more things need to be done, then please explain what's happening in the code comments. I can barely understand it right now. I will review more thoroughly when you cleanup the diff and when I have more time.

 - The country enabling logic should not be intermixed with URL param parsing and region/country validation logic.

an-ko-7:
Hi @ivan-aksamentov. Let me clarify first that for me, this PR and the code so far is only a draft, I intended to submit another PR as a final version, without the diff and with comments, and to delete the current one. Vercel has just been useful for deploying and debugging, since I've had some problems running the project offline.

The link you checked shows the last draft, but it can be a bit hard to understand since I haven't provided comments, indeed. As I'm quite busy I thought of fixing all this in the end.

Also, as I said I don't have extensive experience with React or Next.js, until now I was more into vanilla JS, so sorry if the practises I chose are not ideal.

- I chose the `window` object since I've had some minor trouble with `router`, I didn't imagine `window` would be problematic since it ran with no obvious errors or warnings, but it should be easy to change.
- For the parameters format, as a proof of concept, I chose a style similar to _Our World In Data_, for [example here](https://ourworldindata.org/explorers/coronavirus-data-explorer?zoomToSelection=true&time=2020-03-01..latest&facet=none&pickerSort=desc&pickerMetric=new_deaths_per_million&Metric=Confirmed+deaths&Interval=7-day+rolling+average&Relative+to+Population=true&Align+outbreaks=false&country=IND~USA~GBR~CAN~DEU~FRA). Tildas are used as a delimiter between countries, and I thought of eliminating spaces since I thought that "the space character %20 doesn't look as pretty". Again, these should be easy to change as well if you prefer them in another way.
- For the feature of eliminating invalid countries, I'm not sure if you're suggesting some change.
- The country enabling/disabling logic is not in my last draft yet, I only proposed it. I agree with your description of implementing it.
- For the comment "country enabling logic should not be intermixed with parsing or validation"; again not sure if you're suggesting some change, or just couldn't make sure sure of my implementation due to lack of comments, and made the note regardless. I'm not sure of what an alternative would look like, perhaps I can put some comments in my next version to communicate more easily.

an-ko-7:
Hi @ivan-aksamentov. Let me clarify first that for me, this PR and the code so far is only a draft, I intended to submit another PR as a final version, without the diff and with comments, and to delete the current one. Vercel has just been useful for deploying and debugging, since I've had some problems running the project offline.

The link you checked shows the last draft, but it can be a bit hard to understand since I haven't provided comments, indeed. As I'm quite busy I thought of fixing all this in the end.

Also, as I said I don't have extensive experience with React or Next.js, until now I was more into vanilla JS, so sorry if the practises I chose are not ideal.

- I chose the `window` object since I've had some minor trouble with `router`, I didn't imagine `window` would be problematic since it ran with no obvious errors or warnings, but it should be easy to change.
- For the parameters format, as a proof of concept, I chose a style similar to _Our World In Data_, for [example here](https://ourworldindata.org/explorers/coronavirus-data-explorer?zoomToSelection=true&time=2020-03-01..latest&facet=none&pickerSort=desc&pickerMetric=new_deaths_per_million&Metric=Confirmed+deaths&Interval=7-day+rolling+average&Relative+to+Population=true&Align+outbreaks=false&country=IND~USA~GBR~CAN~DEU~FRA). Tildes are used as a delimiter between countries, and I thought of eliminating spaces since I thought that "the space character %20 doesn't look as pretty". Again, these should be easy to change as well if you prefer them in another way.
- For the feature of eliminating invalid countries, I'm not sure if you're suggesting some change. 
- For the country enabling/disabling logic, perhaps my implementation hasn't been clear due to lack of comments, I thought of it as the most intuitive solution. It's mostly complicated by the feature of elimination, and the enabling logic is a short as possible, just the following:

```
const countriesListRaw = perCountryData.distributions
    .map(({ country }) => (
      countries.includes(country) || loadAllCountries 
        ? { countryName: country, enabled: true } 
        : { countryName: country, enabled: false }
    ))
```
But I will check your proposal again.

ivan-aksamentov:
Closing in favor of https://github.com/hodcroftlab/covariants/pull/279

PR #249, opened 2021-12-21
Use case counts data

### LINKED ISSUE #243: Incorporate Case Count Data
A few people have commented that it could be interesting to incorporate case-count data. I presume we could pull this from Johns Hopkins or OWID? I also assume we'd want a standardizes count as used by OWID (x per million etc). 

Two ways this could be incorporated (either, or both):
- Second Y axis added to Per Country plots and the number of cases overplots the variants (can be turned on/off)
- Multiply together case count & variant frequency to generate a plot showing % variants under the case-count curve (see figure in [this paper](https://www.nature.com/articles/d41586-021-03619-8?utm_source=Nature+Briefing&utm_campaign=17b59a2f87-briefing-dy-20211207&utm_medium=email&utm_term=0_c9dfd39373-17b59a2f87-44960825) - though this is per # sequences). Big caveat here is that we likely only want to make this possible for countries doing a good amount of sequencing. We'd probably want to generate some list of countries that have >=X sequences per case (or X sequences per population) and only allow these transformations on these countries. But for countries with good sequencing, this could be very interesting.

ivan-aksamentov:
Resolves #243 

 - [x] download OWID data
 - [x] transform OWID data into web data
 - [x] display web data


MoiraZuber:
Added a script to combine the two files `data/owid/owid-covid-data.csv` and `web/data/perCountryData.json` into `web/data/perCountryDataCaseCounts.json`. The file has the same structure as `perCountryData.json`, though only the region "World"  is used ("USA" and "Switzerland" are discarded). 

The file contains these new fields for each week & country:
- "percent_counts": For each cluster, percent of sequences (relative to total sequences received)
- "total_cases": Total amount of cases according to OWID
- "percent_total_cases": Fraction of cases sequenced (this could be used to find a threshold for minimum amount of samples sequenced)
- "estimated_cases": For each cluster, estimated number of overall cases by multiplying percent_counts  with total_cases

TODOs:
- [ ] [Line 84](https://github.com/hodcroftlab/covariants/blob/537bb05fd4025c15b04fb1bef57b9ea3ae521073/scripts/include_case_counts.py#L84) Which fields to use on the region level? (Only "World" in our case)
- [ ] [Line 80](https://github.com/hodcroftlab/covariants/blob/537bb05fd4025c15b04fb1bef57b9ea3ae521073/scripts/include_case_counts.py#L80) Which fields to use on the country level? Are some of those I added not necessary? Do we need additional ones?
- [ ] [Line 56](https://github.com/hodcroftlab/covariants/blob/537bb05fd4025c15b04fb1bef57b9ea3ae521073/scripts/include_case_counts.py#L56) What to do if a country is missing from OWID (currently only "Guadeloupe"). Alternative country names can be designated, in two cases I was a little unsure [Line 15](https://github.com/hodcroftlab/covariants/blob/537bb05fd4025c15b04fb1bef57b9ea3ae521073/scripts/include_case_counts.py#L15)
- [ ] Decide on threshold for minimum number of sequences, how to deal with missing OWID data (currently treated as 0 cases)?

emmahodcroft:
This is great progress Moira - thanks! Can you push up a copy of the file that's generated? Since the format is already a tidy JSON we can probably put it in `web/data` directly - but move it if necessary. This way Ivan could take a look and see if it looks like something that can be easily put into a graphical format.

For the questions, not entirely sure I'm understanding, might be best to discuss - but from my interpretation here's what I'd say:
>What fields to use on the region level?

I think we can start with just World. For Switzerland, I think the numbers are too low to try and do cases on a regional level. For US states the case would be stronger, but not sure if it's worth getting into this, either. Particularly since number of cases being sequenced is likely to drop in the coming months. 

>Which fields to use on the country level?

These look good to me from what we discussed and what I'm picturing we need - but here's where Ivan can give us feedback about what's best to have to draw the graphs! (Also we might be able to picture this better when we see the file that's generated.)

> What to do if a country is missing from OWID?

I think if a country is missing from OWID, we just don't generate the data. Without reliable case counts, we can't make any plots. Let's just ensure we handle this gracefully in the code (aka not crash and burn 😆 )

> Decide on threshold for minimum number of sequences

This one I think we might need to tackle in a slightly different way. I think we likely want to make a decision on whether to plot a country at all, or not at all (not have some data points missing for weeks without enough samples, etc). 
I think we may want to make these decisions within your script, and then only print out information for countries that we have decided we want to display - so if we aren't graphing a country, it just doesn't even go into this file. Alternatively, we print out all countries in the file, but each gets a boolean flag `True` `False` for whether to plot or not.

The threshold kind of needs to take two things into account: we want a minimum number of sequences per cases (because if we're only sequencing 10 cases out of 10,000, we have no expectation this is representative). 

But we also perhaps want to take into account countries without much testing - earlier in the pandemic there was a thing going around about some tiny African country was doing the best at sequencing because they had like 10 cases and had sequenced 8 of them (can't remember exact numbers)... but really it seemed like the issue was that they didn't really have testing capacity - they actually had more than 10 cases! So that might imply taking into account something like a minimum cases per population. (Though this may in theory throw out countries like New Zealand who've had very few cases...)

Why don't we first try just a minimum number of cases per sequences, and see what this looks like for the set of countries we have data for on CoVariants. The challenge is, we likely want some kind of running measure which never dips below a certain number, or else we get an alert -- if a country drastically reduces their sequencing, we may want to stop plotting them in the future. We could do something like, from the start of 2021 (let's give everyone 2020 as a freebie), does their seqs/cases ratio dip below X? 
Perhaps you could prepare a little overview of the `percent_total_cases` per 2 weeks so we could get a sense of how this might work - for example, a printout like: (These numbers totally made up, and in %)

```
     wk0    wk2 ...
UK:  1.8    2.0 ...
USA: 2.4    1.9 ...
CH:  0.8    1.3 ...
.
.
.
```
And a little script where we can get some feedback on different thresholds like:
```
At threshold 1.0%:
UK:     Pass for 50/64 2-week-periods
USA:    Pass for 59/64 2-week-periods
CH:     Pass for 34/64 2-week-periods
.
.
.
```

This is very much brainstorming - if you have another idea for how to start getting to grips with what thresholds we want to impose, please try it out!!

emmahodcroft:
Thanks Moira!

@ivan-aksamentov would you mind checking out [this file](https://raw.githubusercontent.com/hodcroftlab/covariants/feat/case-counts/web/data/perCountryDataCaseCounts.json) to see if it seems like the right format for drawing graphs from? We tried to base it off the existing format, since that seems to work for the current graphs 😄 

I asked Moira to include two 'types' of numbers here so we have more options for plotting. With the 'percents' we can plot the whole case numbers and colour in by fraction, like this graph:
image

With the 'case counts' we could offer an option to just show the estimated case counts of each variant as a line, like this from Trevor:
image

Perhaps we could even have a way of toggling between the two? Anyway, I thought it might be easiest to provide both (plus total cases & total sequences) so that we've got a lot of options!

MoiraZuber:
I added a new script `scripts/case_counts_analysis.py`, which produces the output you suggested above. It asks for two input values `threshold` and `period-pass`, which should function the way you described. It can be used to play around with those values to try and find a suitable threshold. @emmahodcroft do you want to give it a try and tell me if it needs any adjustments?

(The first printout you mentioned, listing `percent_total_cases` per week and country, and is produced by the script as a dataframe called `df` and could be plotted if desired, though it is not printed out right now.)

ivan-aksamentov:
I added "Cases" page (`/cases`) and I render the `estimated_cases` field as the plots, similarly how I render `cluster_counts` on the "Per country" page. In fact it's copy-pasted from there pretty much. Similarly, the frequency column in the tooltip is the number of cases of a variant, divided by `total_cases`.

I don't understand what are the other different fields in the new JSON entries:

```json
{
  "week": "2021-10-18",
  "estimated_cases": { "21I (Delta)": 19, "21J (Delta)": 37  },
  "percent_counts": { "21I (Delta)": 0.3333333333333333, "21J (Delta)": 0.6666666666666666 },
  "percent_total_cases": 0.05357142857142857,
  "total_cases": 56,
  "total_sequences": 3,
}
```

I am struggling with Emma's explanation above. Is `percent_counts` different from frequencies as I calculate? 

Why the same thing needs to be rendered as area plot first, and then as line plot in the other case?

Please ELI5. Perhaps more convenient on Slack.

For the current version, see the preview link in the very first message of Vercel bot.

MoiraZuber:
I changed the script so that the missing fields are automatically added to the json. Currently it copies all missing fields directly from `web/data/perCountryData.json`, including max_date and min_date.

emmahodcroft:
Sorry for so long to come back to this

>I am struggling with Emma's explanation above. Is percent_counts different from frequencies as I calculate?

No, hopefully it's not different! I just didn't know if it was easier to supply these pre-calculated.

>Why the same thing needs to be rendered as area plot first, and then as line plot in the other case?
Please ELI5. Perhaps more convenient on Slack.

Yes, maybe let's just stick to the first case first, and we can talk about whether it's worth having another way to show it.

The way you're rendering now is accurate, but basically shows the same as on the Per Country page - so it's not really adding much. What I'd like to show is something like this:
image

Which is, the X axis reflects the total number of cases (I've just realised, we probably need to standardize this so that our X-axes aren't all over the place...) and then the total number of cases (which changes over time) is filled in under the line by the estimated number of cases due to each variant. This way people can see, when waves happened (or when cases were low) - what variants were responsible (or associated)?

@MoiraZuber I've just realised that in order to have the total number of cases be anything like comparable between the different plots, we will need to standardize. On Our World In Data they do this by cases per million people, so we can likely stick to this. Sorry for only thinking of this now! To get this we'll need to add population data for each country. This may already be present in the OWID data - if not I think you can probably get it off the OWID website. We only need to include this once - population will not change fast enough for us to have to worry about this.
Could you then use this to standardize the total number of cases & estimated number of cases per million? I'm on the fence about whether we should include the originals as well (make these new entries) or overwrite the ones we have now (unstandardized) and just change the variable name. In an ideal world we'd have both - people will find that raw-data useful. But I'm a little concerned how big the file might end up getting...

It sounds like Ivan is happy to calculate the frequencies within the plot, so one option would be to drop including the `percent_counts` and replace with `stand_est_cases` instead, if @ivan-aksamentov can confirm this is ok! Then we just add a `stand_total_cases` as well.

emmahodcroft:
>I added a new script scripts/case_counts_analysis.py, which produces the output you suggested above. It asks for two input values threshold and period-pass, which should function the way you described. It can be used to play around with those values to try and find a suitable threshold. @emmahodcroft do you want to give it a try and tell me if it needs any adjustments?

This is great Moira - thank you! It'll be a little tricky to keep an eye on this going forward -- as countries stop sequencing as much we'll have to revaluate what we want to keep plotting.

I think at the moment, after playing around a little, I'd vote we start with a 3% sequencing threshold for 50% of the 2 week periods we track.
Can you add a little something to the script, so that each country has a `plot` variable (or similar) that can be `True` or `False` and set it based on this?

It would be good to make it pretty easy to change the two thresholds in the script (somewhere not to buried!) so that we can adjust them as needed over time :)

ivan-aksamentov:
@emmahodcroft 
> What I'd like to show is [...], the X axis reflects the total number of cases [...] is filled in under the line by the estimated number of cases due to each variant

I pushed a new version. Something like this?




Omicron makes older things not very useful with linear Y axis. I could not make the log scale to work properly with stacking (at least in reasonable time). Without stacking it's super ugly:



Perhaps normalizing by population can make peaks less pronounced and this will make plots with linear axis more useful?

ivan-aksamentov:
> We'll want to only plot countries where plot is True [..] I figured adding this variable to the file would be the easiest way for you - does that seem so?

The easiest for web is to not emit stuff into JSON. It will disappear automatically.Why would user download data that is not shown?
In case of a boolean, it will need programmatic intervention. So if there's no good reason for keeping them to be in the web JSON, then just not emit them.

ivan-aksamentov:
> To get this we'll need to add population data for each country. This may already be present in the OWID data

If I remember correctly, the OWID data that we download already has the cases normalized by population. But I haven't checked.

emmahodcroft:
>I pushed a new version. Something like this?

Yes - wonderful! That's exactly the type of plot I meant! :) I agree though, with the current numbers the y-axis is too variable.

>If I remember correctly, the OWID data that we download already has the column for cases normalized by population

Ivan - you're right! And @MoiraZuber is already using it - which is great, but also unfortunate as this means if we want to dampen Omicron we'll have to get into something like logging, which sounds like it's not really working at the moment. 

I actually think even the current plots are not so bad - yes Omicron is big, but you can still see how much waves were composed of variants. But, might not be ideal.

I'm not 100% sure of the difference between 'line' plots (like Per Variants) and the 'area' plots you use here (and on Per Country) - but one option might be to have a way of switching the plotting. You can have unlogged stacked area plots (exactly what you have above) - or (toggle switch) - line plots that are logged and unstacked (which I think is fine for line plots). Would that be a thing? Or too complex?

MoiraZuber:
I added the threshold and removed all countries from the output that don't pass, like @ivan-aksamentov suggested. 

Concerning the normalized data, we do download it but I don't think I use it as input yet? We get `total_cases, new_cases` as well as `total_cases_per_million, new_cases_per_million`. Currently the former is used. It would be pretty easy to switch to the "per_million" entries, if that would solve the problem?

ivan-aksamentov:
@emmahodcroft 
> You can have unlogged stacked area plots (exactly what you have above) - or [toggle switch] - line plots that are logged and unstacked (which I think is fine for line plots). Would that be a thing? Or too complex?

That will be some work, yes. Not astronomically hard. I'd say we add it one of the later releases perhaps.  

---

@MoiraZuber 
> I added the threshold and removed all countries from the output that don't pass, like @ivan-aksamentov suggested.

Great! The changes should be reflected in the Vercel deployment automagically (on every commit). Is everything looking correctly there?


>  It would be pretty easy to switch to the "per_million" entries, if that would solve the problem?

I believe is that what's needed. But let's see what Emma says. You can try to adjust Python code an regenerate the JSON (without changing its format) and see how it looks in the real deployment.

The Vercel bot's first message shows the latest deployment, and  "View deployment" buttons against every commit show historical deployments, for every commit, as were working on the PR over time, so you can compare multiple versions if needed.

ivan-aksamentov:
> I actually think even the current plots are not so bad - yes Omicron is big, but you can still see how much waves were composed of variants. But, might not be ideal.

Another solution is to have plots implementations that are actually make sense - better layout, size, pan & zoom and other quality of life features. You could for example scroll left and zoom to see the historical clades better. The plots that Richard Goater and his team have implemented could really save the day.

emmahodcroft:
>I added the threshold and removed all countries from the output that don't pass, like @ivan-aksamentov suggested.

Thank you Moira! I can indeed see the changes reflected, as Ivan says. 

>Concerning the normalized data, we do download it but I don't think I use it as input yet? We get total_cases, new_cases as well as total_cases_per_million, new_cases_per_million. Currently the former is used. It would be pretty easy to switch to the "per_million" entries, if that would solve the problem?

Oh! Perhaps I was reading the code too fast (probably so). I thought that I saw you were already using cases per million. If not, let's try switching to that!
Ivan, if we keep this, we'll need to adjust the y-axis labels to reflect that it's per million. (I'm hoping that's not too hard)

>That will be some work, yes. Not astronomically hard. I'd say we add it in one of the later releases perhaps.
Also, the switching will be insanely slow, same as switching pages currently, because many plots need to be rerendered from scratch. So the user experience may not be fantastic.

Ok, I was hoping maybe this was something that might be an easier implementation, but if not, then yes, let's not hope for that yet - let's in particular see what things look like when we switch to cases per million. 


>Another solution is to have plots implementations that actually make sense - better layout, size, pan & zoom and other quality of life features. You could for example scroll left and zoom to see the historical clades better. The plots that Richard Goater and his team have implemented could really save the day.

This would be fantastic, to have interactive plots - but I don't know who might be able to make this happen - it seems like it might be quite a heavy lift...

MoiraZuber:
I removed `percent-counts`from the output and switched to using `per_million` input (except for `percent_total_cases` and the threshold check, there we still use the unstandardized counts).

emmahodcroft:
Thanks @MoiraZuber ! Can you check if the below should be correct:

@ivan-aksamentov Could you switch the visualization scripts from using `total_cases` and `estimated_cases` to `stand_total_cases` and `stand_estimated_cases`? Then we should be able to see how this compares :)

ivan-aksamentov:
What's the point of renaming? Can we just keep the old variable names to avoid unnecessary work? I will work automagically then without my intervention.

emmahodcroft:
In case people are accessing the data raw, I think it's good to have it clear(er?) that these are standardized values, not absolute. But I also take the point that these names aren't super clear. Since you can't comment in JSONS, I guess there isn't another way to let people know this in the file - or is there?

ivan-aksamentov:
Alright, I renamed variables in the app.

I only use these:

```
week
stand_total_cases
stand_estimated_cases
```

I don't know what the rest of them are and whether I should be using them. If not, it's better to remove them to avoid users downloading unused extra data on every visit. The thresholds or other data, if used only internally should not be leaking into the app.

ivan-aksamentov:
What does the data point like this mean?

```json
"stand_estimated_cases": {
  "20A/S:439K": 0,
  "21A (Delta)": 0,
  "21I (Delta)": 0,
  "21J (Delta)": 0
},
```

What is the difference between this and not having a data point? What is the difference between 0 and no value for a cluster? Related: https://github.com/hodcroftlab/covariants/pull/255#issuecomment-1066956779

If zeros are not significant, then we should probably try to remove them and see how this behaves. It will make JSON files like 50% smaller, web app a bit faster, and will slow down the catastrophic git repo growth we are observing. I also suspect these 0 vs missing might be programming or numerical bugs on Python side.

ivan-aksamentov:
> However, unless there's other ideas for plotting, it seems like something we might just have to live with?

The idea is to eventually either have a "log scale" toggle, or a more interactive plot implementation (with pan&zoom), or both. But I currently cannot promise that I can implement either in the short-term unfortunately.

ivan-aksamentov:
I assume we also need the filtering sidebar? Or perhaps these plots will get merged to another page? 

If omicron checkbox is unchecked it can be done such that the Y axis will rescale to fit the other variants, making them more pronounced.

emmahodcroft:
>What is the difference between this and not having a data point? What is the difference between 0 and no value for a cluster? Related: https://github.com/hodcroftlab/covariants/pull/255#issuecomment-1066956779

This is the same situation as in the other situation - if we have a zero, that means we have no sequences (no data), but since we can only detect a variant by observing it, we interpret this has having no variant/no cases from that variant.
We could try removing zero values from outputting, that would be nice if it would shrink files -- are we sure the web side of the code will be ok with it? How will it behave if that means for that entire time period there are then no entries? Do we remove that time period entirely?

>The idea is to eventually either have a "log scale" toggle, or a more interactive plot implementation (with pan&zoom), or both. But I currently cannot promise that I can implement either in the short-term unfortunately.

We might have to settle on this for the moment, and see if we can improve the graphing overall in the future.

>Another solution would be to stop making the viruses that cause peaks like this. I'd take this one, because it does not require any programming from my side.

Full agree, but unfortunately outside of my control... 🙃 

>I assume we also need the filtering sidebar? Or perhaps these plots will get merged to another page? (which already has the sidebar)

Filtering sidebar would be nice. I think it's good to keep this page separate, since we don't show the same number of plots here as we do on other pages. The interpretation is also a little different, since these are 'estimated' case numbers not real values based on number of sequences.
I'm of two minds on whether Y-axis should resize. On the one hand, this would be very useful to provide a 'faux-zoom' for people interested at looking at previous peaks in more detail. On the other hand, as you say, perhaps questionable to allow total number of cases to be 'decreased'. But perhaps allowing a way to let people get a better look at previous peaks - since we have no better option to improve the 'viewing' right now - is more important. So I lean towards yes, letting y-axis resize.

ivan-aksamentov:
> are we sure the web side of the code will be ok with it? How will it behave if that means for that entire time period there are then no entries? Do we remove that time period entirely?

Yes, web will be fine. Many variants already don't have values, others have 0's. Missing single variant value means there is no little dot marker displayed when you mouseover. If the entire objects (for a given week) is missing, then mouse just jumps over this week (the interval is larger). Having a zero means that the little dot will be at 0 and the whole curve goes to 0.

emmahodcroft:
Ok, happy to try this. Moira, could you have a go and removing entries where the value is 0? We are having an issue with the dataset size - I've just had to restart building master twice on CircleCI, and we think it fails because of the size 😕 

@ivan-aksamentov - I'm testing this for the Per Country data in #291 - doing Per Variant is a bit more complicated.

emmahodcroft:
Actually - from testing this on Per Country page (#291) the plotting doesn't seem to handle not having zeros very well. Until we know we can get that to work (and know exactly how to format so that it does work), let's hold off on removing zeros.

The CircleCI builds are taking a lot of hand holding so it would be good to tackle this, but we need to know the plots will look the same in the end.

ivan-aksamentov:
I removed the code from the web app that erased zero values from cases data. This was needed when I played with log scale Y axis. Plots break on log(0), so 0s needed to be removed.

Similarly to how it happens in https://github.com/hodcroftlab/covariants/pull/291 removing zeroes creates white holes in the plots.

With zeroes everything is smooth again.

MoiraZuber:
Updated case count data. Currently with data from master that is >10 days old, will try to merge new data in later.

ivan-aksamentov:
I made the following changes:
 - [x] add filtering sidebar to cases page
 - [x] add label for y-axis to cases page plots ("cases per million")

emmahodcroft:
I've incorporated downloading OWID cases and running the case script into `make web-data`. This should allow easy updating as part of the pipeline.

emmahodcroft:
The filtering sidecar works great @ivan-aksamentov - and looks great! As do the y-axis labels! 😁

emmahodcroft:
I've merge in master and updated the cases - hopefully this should show something that's quite up-to-date...

ivan-aksamentov:
One thing we forgot to discuss is how to name the navbar link (URL), what text should be there and where to put it. Right now it's called "Cases", leads to `/cases` URL, and it's between "Variants" and "Per Country". Perhaps it needs to be after "Per Variant", as less important?

The order, URL and text can be tweaked here:
https://github.com/hodcroftlab/covariants/blob/23e447ca7f96053718f1266830912ba30f7e5b75/web/src/components/Layout/NavigationBar.tsx#L27

emmahodcroft:
Thanks Ivan, good catch! I think 'cases' (name and URL) is great. But agree that moving after Per variant is a good idea. I'll do that now (along with another minor page header edit).

I'm hopeful this is pretty much ready to go after this.

The best way to do a full test on this is to do a full run, but unfortunately since we're now running in Scicore, that means (as far as I know) we can't do the web-page display locally, therefore you have to push The Whole Damn Thing (TWDT) up somewhere to see the final result. This is slightly risky as if you have to revert, it's a bit of a mess with the new clades files etc.

However, I guess one has to start taking risks somewhere when doing a big merge, I guess?
Unless there's a thought of a better way - I'll plan to use _this branch_ for the usual update, and then push TWDT (cluster updates and all) back to this branch. It'll deploy, and we can check it. If all looks good, I'll write a twitter thread and merge it in 😄

ivan-aksamentov:
Totally! Just YOLO it!

PR #251, opened 2021-12-30
Provide (more) consistent linting and formatting with black and flake8

### LINKED ISSUE #245: Consistent python style, linting and formatting for `scripts` folder
Currently the python scripts in the `./scripts/` folder do not seem to strictly follow any linting, formatting, typing or style guide, docstrings are often missing and no configuration files to ensure code style consistency across IDEs seem to be present.

To enable consistent collaboration, it might perhaps be of interest to setup:

- Linting, e.g. with [flake8](https://github.com/pycqa/flake8) or pylint
- Formatting, e.g. wtih [black](https://github.com/psf/black)
- Import order sorting, e.g. with [isort](https://github.com/PyCQA/isort)
- Enforcing of docstrings, e.g. with [flake8-docstrings](https://github.com/pycqa/flake8-docstrings) and [Google Style Python Docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)
- Enforcing consistent styles before commit with a [pre-commit hook](https://pre-commit.com/)
- Testing, for example with [pytest](https://docs.pytest.org/)
- Linting as part of the CI/CD pipeline

maltekuehl:
Resolves #245.

The goal of this PR is to provide the necessary basis to facilitate contribution from different people to the python code in the future while at the same time keeping disruption to the existing workflow and changes to a minimum.

Features of this PR:
- Adds a `requirements.txt` to the `scripts` folder to keep track of dependencies
- Provides a `Makefile` to easily install the necessary dependencies and setup linting and formatting
- Uses flake8 and black for linting and formatting inside the `scripts`
- Adds most of the current linting problems to the `.flake8` ignore list with a description of the problem. This prevents yellow and red flags from popping up all over the code in the IDE but allows for progressive removal of the flags and enhancement of the code style if desired.
- Reformats the Python files using this setup
- Includes the VS Code settings to enable linting and formatting

I've tried to do justice to most of the points raised by @ivan-aksamentov in his [comment](https://github.com/hodcroftlab/covariants/issues/245#issuecomment-998104731) on #245 and tried to keep this as light and unobtrusive as possible but I'm aware that this is an opinionated solution that may require changes.

Debatable point may include:
- Should the VS Code settings be included in `/scripts/.vscode/settings.json` since this is an IDE-specific feature?
- Use of virtual environments for the setup
- Use of second `Makefile` for setup: The reason behind this is to make linting and formatting completely optional and only part of the `scripts` folder. However other solutions might be providing a bash script for setup or integrating the `make` endpoint into the project's main `Makefile`.
- Some imports from the code are not in the `scripts` folder, their paths could be added to `python.analysis.extraPaths` in the VS Code `settings.json`.
- The `requirements.txt` file currently only contains the package names but not their versions, perhaps the output of `pip freeze` could be added here to add exact versioning. 

Due to missing data it is also not possible for me to test all of the functions. Since only the formatting has been changed their functionality should not be affected but it would probably be good to test and confirm this anyway.

Happy to hear your feedback and to either move forward with this draft or reconsider the approach or certain aspects.

### DIFF
--- a/scripts/.flake8
+++ b/scripts/.flake8
@@ -1,0 +1,17 @@
+[flake8]
+ignore = E203, # Whitespace before :
+    E266, # Too many leading hashes in comment
+    E402, # Module import should be at the top
+    E501, # Line length exceeds max-line-length
+    E712, # Comparison to boolean value should use is
+    F401, # Imported but unsued
+    F403, # Unable to detect import names
+    F405, # Value may be undefined or imported via star import
+    F601, # Repeated dictionary key
+    F632, # Use ==/!= to compare constant literals
+    F821, # Undefined names
+    F841, # Local variable is assigned but never used
+    W605, # Invalid escape sequence, "\d" in regular expression
+max-complexity = 75
+max-line-length = 120
+exclude = .git,__pycache__,.mypycache,.venv
--- a/scripts/.gitignore
+++ b/scripts/.gitignore
@@ -1,0 +1,3 @@
+.venv
+.mypy_cache
+requirements.lock.txt
--- a/scripts/.vscode/settings.json
+++ b/scripts/.vscode/settings.json
@@ -1,0 +1,28 @@
+{
+    "python.defaultInterpreterPath": ".venv/bin/python",
+    "python.formatting.provider": "black",
+    "python.formatting.blackPath": ".venv/bin/black",
+    "python.formatting.blackArgs": [
+        "--line-length",
+        "120",
+    ],
+    "python.linting.pylintEnabled": false,
+    "python.linting.flake8Enabled": true,
+    "python.linting.flake8Path": ".venv/bin/flake8",
+    "python.linting.mypyEnabled": false,
+    "python.linting.mypyPath": ".venv/bin/mypy",
+    "python.linting.enabled": true,
+    "python.analysis.extraPaths": [ // setting for pylance when used
+        "PATH_TO_MISSING_MODULES",
+    ],
+    "[python]": {
+        "editor.formatOnPaste": false, // not supported by black
+        "editor.formatOnSaveMode": "file"
+    },
+    "editor.formatOnSave": true,
+    "editor.rulers": [
+        {
+            "column": 120,
+        }
+    ]
+}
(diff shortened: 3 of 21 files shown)


ivan-aksamentov:
Hi Malte @maltekuehl,

Thanks for looking into this. I like it!

I address the "Debatable points":

 - > Should the VS Code settings be included [...]

   I don't have opinion on vscode config. I use PyCharm, so as long as it does not bother Emma, I am fine with it. I'd say let's give it a try.

 - > Use of virtual environments for the setup

   Same. As far as I understand, deps can also be installed through 

   ```bash
   pip3 install -r requirements.txt
   ```
   , so any environments are optional and they don't ask for money, so I'd say we keep it if it's useful for someone.

 - > Use of second Makefile for setup

   Yeah, I'd prefer the simplest solution with the single makefile at the root.
   The scripts folder is somewhat cluttered already. Could you indeed move all the config files to the root of the project, so that they don't distract Emma too much?

 - > Some imports from the code are not in the scripts folder, their paths could be added to python.analysis.extraPaths in the VS Code settings.json.

   Which imports do you have in mind? Is there a way to address this in the linter config, for cases when VSCode is not used?

 - > The requirements.txt file currently only contains the package names but not their versions, perhaps the output of pip freeze could be added here to add exact versioning.
 
   Not sure if we have any version constraints. These packages are fairly common and stable, sort of like Python's standard library. I'd say we keep them on latest unless proven that a freeze is needed.


Additionally:

 - Could you add makefile targets for each of the tools so that they can be invoked easily? For example:

   ```bash
   make lint
   make lint-fix    # to apply auto-fixes, if any
   make format   # to fix code formatting
   ...
   ```

 - I suggest we revert all the code changes for now. We would then merge only the config files, and synchronize with Emma, telling her what commands to run to apply the fixes. This is to avoid messy merge conflicts, because Emma keeps modifying the scripts constantly and it may also break some of the changes she might have in branches and stashes. The easiest way to revert is to pick files from the parent commit of this branch, copy them over to a safe place check out the tip of the branch again and then copy them back. This should remove these files from GitHub diff and resolve the existing conflict. 

   Emma will then reapply the formatting fixes when it's convenient for her. And after formatting is fixed, we can proceed to picking on linting problems in small chunks. I'd like to decouple formatting and fixes, to make sure that any potential changes in code behavior can be spotted.


Really appreciate your contribution!

PR #254, opened 2022-01-04
Add Variant Graph to Variant Intro Page

emmahodcroft:
Add a picture of the Nextclade Clade graph to the Variant intro page.

However, this isn't ready to merge as the image size isn't looking great. Tried with HTML but this messes up the indent to the right. With Markdown doesn't mess up indent, but couldn't figure out how to scale.
image


emmahodcroft:
Thank you for working on this @ivan-aksamentov ! The deploy looks great to me, at multiple resolutions. I'm happy with this if you are!

PR #255, opened 2022-01-05
Add "lesser than cutoff" value to plot tooltip

osirisinferi:
Currently, when I'm looking at the `/per-variant` page and a graph has a value (i.e.: a line of the graph just barely visible but not nothing), but below the cutoff value defined in `ClusterDistributionPlotTooltip.tsx` (const `EPSILON`), the tooltip will display "`-`" as value.

To me, this doesn't make much sense, as I would think "`-`" means "no data" or "nothing". But I can clearly see a tiny little bit of graph. Not nothing.

Therefore, I suggest changing the tooltip code to display "`<0.01`" (dynamically based on `EPSILON`) if a value is present, but below the cutoff value.

Note that ESLint will raise an "unicorn/no-nested-ternary: Do not nest ternary expressions." warning.. But I don't have any experience with TypeScript to be honest and a lot of pages on the world wide web say just to disable the `no-nested-ternary` rule :roll_eyes:

### DIFF
--- a/web/src/components/ClusterDistribution/ClusterDistributionPlotTooltip.tsx
+++ b/web/src/components/ClusterDistribution/ClusterDistributionPlotTooltip.tsx
@@ -115,7 +115,11 @@ export function ClusterDistributionPlotTooltip(props: ClusterDistributionPlotToo
                 </td>
                 <td>{interpolated && '*'}</td>
                 <td className="px-2 text-right">
-                  {value !== undefined && value > EPSILON ? formatProportion(value) : '-'}
+                  {value !== undefined && value > EPSILON
+                    ? formatProportion(value)
+                    : value !== undefined && value !== 0
+                    ? `<${EPSILON}`
+                    : '-'}
                 </td>
               </tr>
             )


ivan-aksamentov:
Hi @osirisinferi,

Thanks so much for looking into this. Indeed, the "-" sign does not mean anything and we need to be more precise.

I think that there are 3 cases:
 - the data is present and number is within min precision without being all-zeroes
 - the data is present, but the number is smaller than the minimal precision we can display without being all-zeroes
 - data is missing (corresponds to missing field or null in JSON data)

Same for the plots on /per-country page and perhaps other places where this comes up. I think the logic there is different and it shows "0.0" for the case 2.

Emma, @emmahodcroft what's the desired presentation here?

As an additional improvement, if this is scientifically sound, we can show percentages instead of fractions (i.e. multiply the value by 100 and append a % sign). This will allow to fit more information in the same amount of screen space, as there will be no need for `0.`, so we can display 3 significant digits instead of 2 fr most values. Or even more if necessary, and make tooltips a bit wider.

For the case of missing data I imagine also saying explicitly "n/a", "no data" or similar.

We can play with different options and see what's looks the best and the most informative.

osirisinferi:
> With regards to ternary expressions, I think eslint is right. Me, personally, I am having difficulties understanding the current logic in this snippet. This code should probably go into the formatting function's body or into a separate function, and should be normal conditionals instead of ternary.

Well, a (nested) ternary was the only thing I managed to put between the `{` and `}` 🙄 An entire `if .. else if .. else` block didn't work 😛 I mostly code in Python, so TypeScript is fairly new to me..

ivan-aksamentov:
@osirisinferi No worries about Typescript. I'll move this logic to a function outside of react component code, so that normal, more comprehensible code can be written. This will also make things faster, because right now this code runs on every render, whether values change or not. Just want to understand first what needs to be done conceptually. I will wait for inputs from Emma.

emmahodcroft:
Sorry for missing this for so long Ivan!

>As an additional improvement, if this is scientifically sound, we can show percentages instead of fractions (i.e. multiply the value by 100 and append a % sign). This will allow to fit more information in the same amount of screen space, as there will be no need for 0., so we can display 3 significant digits instead of 2 for most values. Or even more if necessary, and make tooltips a bit wider.

Yes, it would be absolutely fine to switch to "%" and show 3 digits, and happy for you to do this if it saves space & allows for better presentation.

> For the case of missing data I imagine also saying explicitly "n/a", "no data" or similar. Again, even in this case, the "-" sign have no universal meaning, so we probably should be more explicit.

I think this situation may differ between the Per Country and Per Variant, so I'm going to work through them both separately and see what's the same & what's different...

>3. data is missing, which corresponds to the field begin missing or null in JSON data
>4. the data is present and the number is exactly 0

One thing with the data we're working with is that 'exactly 0' and 'no data' are the same - we can only detect a variant if we get a sequence of it, and if there are no sequences, that's both no data and means that we would say there is 0% of this variant. At the moment at least on the Per Country page these are values plotted with '-' and I think changing these to not show at all (be absent from the tooltip) would be one way around this entirely - then that variant is absent in the same way that any sequences for it were absent - thus we say nothing about it (implying 0%). 

On Per Country, very small values do sometimes show up as 0.00 and I agree this can be confusing - you can tell the difference because they'll have a non-zero number in `Num seq` -- from 24 Jan 2022 USA, `20I (Delta)` has 11 `Num seq` for a `Freq` of `0.00`. It would be nice to change this to something like `<1%`, which differentiates from "true exactly 0" (covered above).

**For Per Variant** things get a little more confusing in all directions, because:
1) I generate smoothed data, meaning we aren't dealing with real values anymore
2) I believe the plot then further smooths between points (Ivan, is this right?)

I think this is what leads to the situation where the tooltip can show `-` but the line is still visible:
image
In at least some cases it seems that the tooltip shows `-` when a value is present but very small - above, as far as I can tell (??) Thailand does have a value for this point: `0.009881...` . But this seems to be different behaviour from Per Country, where as far as I was able to tell, even if the value is very low, it will show `0.0` rather than `-` (only showing `-` if it's true zero/no data). 

Looking at the above data point, as far as I can tell, Thailand, Canada, UK, and USA all have non-zero (but very small) values, but display `-`. From Philippines onward, the value is `0.0`

So, I think taking a similar approach here would work actually - switch to %, show small numbers as <1%, and remove showing countries where we have no data (0.0 or that country name doesn't exist at this time point).

For the moment the interpretation of this may remain a little tough, because the lines may still be a little visible even if the value is 0 (if they are smoothed), and the smoothed data I generate means we aren't truly looking at the truth.... but it's on the to-do list to change Per Variant to work like Per Country (real numbers, not smoothed data) which will make this all clearer!

emmahodcroft:
From this PR I think there's two take-aways:

1. On Per Country, let's try not to display anything when values are 0 or -, which in this case are identical. This should be resolved by moving to the new plotting library in #319  (at least as it displays currently)
2. Removing the smoothing from Per Variant plots may go some way to resolving the weird behaviour of the small values/zeros. This is part of #295 and I'll add a reminder there to check this before we merge. If it persists, let's make another issue to fix this after switching to 2-week-no-smoothing plots.

With both of these in mind, it seems best to close this for now, and perhaps draw on this PR if we need, if the issue persists after getting #295 working?
If either @ivan-aksamentov or @osirisinferi disagree, please reopen!

PR #256, opened 2022-01-09
use router query string to preselect region and variants

scko823:
this is related to https://github.com/hodcroftlab/covariants/issues/30

## What
In this PR, I add logic to do the following
0. refactor so the `region` React state is now using an enum instead of string
1. look for the `region` query param to compute initial React state for the region
2. when the user switch region, it automatically update the query param
3. look for the `variants` query param to compute initial React state for the `ClusterState` (please note that user selecting/deselecting a variant will not update the query param)

## Motivation
This PR is very much a beginning PR to tackle https://github.com/hodcroftlab/covariants/issues/30
Due the complexity of the React states (regions, clusters) and the complexity of the world data (continents vs countries/places) I decided this PR is a good stopping point for me to show a demo of what I have.

But generally, the idea is to use the query params as the `source of truth` and any update on the UI checkboxes should be update the router's query params

Future `TODOs` will be
1. add the ability to allow user to select states when region is US
2. add the ability to allow user to select regions when region is Switzerland
3. add the ability to update the variants query params when a different variant is selected/deselected
4. add the ability to update the query params when a different country/continents is selected/deselected

## Demo

https://covariants-git-fork-scko823-query-string-country-hodcroftlab.vercel.app/per-country?variants=21J%20(Delta)&variants=21K%20(Omicron)&region=switzerland

https://covariants-git-fork-scko823-query-string-country-hodcroftlab.vercel.app/per-country?variants=21J%20(Delta)&variants=21K%20(Omicron)&region=usa

### DIFF
--- a/web/src/components/CountryDistribution/CountryDistributionPage.tsx
+++ b/web/src/components/CountryDistribution/CountryDistributionPage.tsx
@@ -1,4 +1,6 @@
 import { mapValues } from 'lodash'
+import { stringify } from 'querystring'
+import { useRouter } from 'next/router'
 import React, { useCallback, useMemo } from 'react'
 import { Col, Row } from 'reactstrap'
 
@@ -23,6 +25,7 @@ import {
 
 import { CountryDistributionPlotCard } from './CountryDistributionPlotCard'
 import { useCountryAndClusterQuery } from './useCountryQuery'
+import { getCurrentQs } from './utils'
 import { CountryFlag } from '../Common/CountryFlag'
 import { USStateCode } from '../Common/USStateCode'
 import { PageHeading } from '../Common/PageHeading'
@@ -31,8 +34,9 @@ const enabledFilters = ['clusters', 'countriesWithIcons']
 const { regionNames, regionsHaveData } = getRegions()
 
 export function CountryDistributionPage() {
+  const router = useRouter()
   const {
-    state: { region, setRegion, setPlaces, places, countryDistributions, currentClusters, setClusters },
+    state: { region, setPlaces, places, countryDistributions, currentClusters, setClusters },
   } = useCountryAndClusterQuery()
 
   const regionsTitle = useMemo(() => (region === Region.WORLD ? 'Countries' : 'Regions'), [region])
@@ -107,6 +111,27 @@ export function CountryDistributionPage() {
     const contentFilename = getPerCountryIntroContentFilename(region)
     return getRegionPerCountryContent(contentFilename)
   }, [region])
+
+  const setRegion = useCallback(
+    (nextRegion: Region) => {
+      if (region === nextRegion) {
+        return
+      }
+      const fullPath = `${router.basePath}${router.pathname}`
+      const nextRegionQs = { ...getCurrentQs(router) }
+
+      if (nextRegion === Region.WORLD) {
+        delete nextRegionQs.countries
+      } else if (nextRegion === Region.UNITED_STATES) {
+        nextRegionQs.countries = 'usa'
+      } else if (nextRegion === Region.SWITZERLAND) {
+        nextRegionQs.countries = 'switzerland'
+      }
+      // eslint-disable-next-line @typescript-eslint/no-floating-promises
+      router.replace(`${fullPath}?${stringify(nextRegionQs)}`)
+    },
+    [region, router],
+  )
 
   return (
     <Layout wide>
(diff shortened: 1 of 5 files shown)


web/src/components/CountryDistribution/CountryDistributionPage.tsx
```diff
@@ -10,53 +12,48 @@ import { DistributionSidebar } from 'src/components/DistributionSidebar/Distribu
...
   const iconComponent = useMemo(() => {
-    if (currentRegion === 'World') return CountryFlag
-    if (currentRegion === 'United States') return USStateCode
+    if (region === Region.World) return CountryFlag
+    if (region === Region.UnitedStates) return USStateCode
     return undefined
-  }, [currentRegion])
+  }, [region])
+
+  const { enabledClusters, withClustersFiltered } = useMemo(() => {
```
scko823, line 49: I combined the two different `useMemo`s into a single one

web/src/components/CountryDistribution/useCountryQuery.tsx
```diff
@@ -0,0 +1,95 @@
...
+    return queryString
+  }
+  return []
+}
+
+/**
+ * A react hook that take in router query params as source of truth
+ * and compute the React component states for UI display
+ */
+export const useRouterQuery = (): {
```
scko823, line 26: main part of the PR. This hook abstract the previous logic, and also add logic to take in router's query params as source of truth, and use it to compute `region` (previously named `currentRegion`), `places`, `currentClusters`

web/src/components/CountryDistribution/CountryDistributionPage.tsx
```diff
@@ -110,9 +108,30 @@ export function CountryDistributionPage() {
   }, [setPlaces])
 
   const IntroContent = useMemo(() => {
-    const contentFilename = getPerCountryIntroContentFilename(currentRegion)
+    const contentFilename = getPerCountryIntroContentFilename(region)
     return getRegionPerCountryContent(contentFilename)
-  }, [currentRegion])
+  }, [region])
+
+  const setRegion = useCallback(
```
scko823, line 115: instead of directly mutating the React `region` state, this will instead just mutate the query params. Once it is updated, the React state will recompute accordingly

emmahodcroft:
Hi @scko823 I'm really sorry to be so late looking at this - I played around with the demo and it looks like it's working well so far! It's very responsive and successfully modified the URL as I selected the different regions.

This would be a fantastic feature to implement more fully - I didn't know if you had had a chance to work on any of the other TODOs you listed?

scko823:
hey @emmahodcroft thank you for replying and the feedback! I have not been working on it since I submitted this PR, because I don't want to invest too much time on it before I get any feedback. But sound like this is solving/improving problem, and I am happy to resume working on it.

> This would be a fantastic feature to implement more fully - I didn't know if you had had a chance to work on any of the other TODOs you listed?

No, but in my mind, each of the TODO can be its own branch/PR; in fact I think the next feature to work on will be `add the ability to update the variants query params when a different variant is selected/deselected` as half of this is actually implemented in this PR.

ivan-aksamentov:
Hi @scko823,

Thanks so much for looking into it.

I liked the idea to move the state around and to refactor things. However, overall the implementation based on useState() and useEffect() where all places are squeezed into one large object, seems to be rather verbose and requires a lot of boilerplate code to synchronize things. 

These thoughts prompted me to try more centralized approach for state management: https://github.com/hodcroftlab/covariants/pull/279 and it was super easy and straighforward, and even reduced the amount of code to maintain overall.

So I will close this PR in favor of https://github.com/hodcroftlab/covariants/pull/279. However, I wanted to give you at least a little bit of credit for the ideas and work, so I based my PR on top of your branch, to keep some of your commits.

Thanks again for your work!

emmahodcroft:
Thanks again for working on this @scko823 and I'm sorry about the change in direction. However, revisiting this led to rethinking it, and as Ivan said, some of your contributions and ideas were still really valuable! I'd still be keen to ensure you're recognised when we can make this public - do you have a twitter account I can tag?

scko823:
hi @ivan-aksamentov and @emmahodcroft thank you for the comment! overall I agree that managing this via a state management system is also a valid approach to problem, and I am glad to have either helped or inspired #279. Look forward to seeing that to be released

Thank you for the offer to do public shoutout, I don't really have a twitter present and I don't think it is necessary 😄 At the same time, I will be looking into other ways to contribute to the codebase over the weekend. Great job putting the site together and making the data available and presenting it in a digestible way for the public. Thank you!

PR #259, opened 2022-01-11
Adjust wording on clade figure slightly

emmahodcroft:
Just adjusts the wording underneath the clade figure slightly.

@ivan-aksamentov I wanted to also add a link to https://github.com/nextstrain/ncov-clades-schema in there, probably around "Nextstrain clades" - so people can fully see where the figure comes from - but I couldn't figure out how to get in-text links working in the TSX file.

No worries if this is a thing that's a pain, we can live without - was just thinking might be helpful for others who might want to use the figure!


emmahodcroft:
I'm going to go ahead and merge this in since it's such a small change (I made a branch because was trying to do the URL). We can make another branch to try that again if we want. This is about as non-urgent as things come.

PR #267, opened 2022-01-22
Update S.K417.md

jacquesgautrais:
Correct for which Variants the mutation is in.

(github has "detected" a change for line 4, but I did not touch it, and I do not see any difference actually...)

Thanks for covariants.org :-)


emmahodcroft:
Thank you @jacquesgautrais ! Yes, I'm not sure what's going in line 4 either. Could be a line-ending change - I sometimes edit in Windows. Either way, doesn't look like it changes anything in reality! 
Do you have a twitter to credit?

PR #274, opened 2022-02-11
Get mutations from metadata

MoiraZuber:
Remove the need to download `mutation_summary_gisaid.tsv` by taking `deletions` and `substitutions` directly from `metadata.tsv` columns.

**TODO:** 
- [ ] [Line 33](https://github.com/hodcroftlab/covariants/blob/8afe2d8e5342060b7e68639d04dbed79e5d752c3/scripts/allClusterDynamics_faster.py#L333) Do we need to remove sequences from meta which don't have substitutions and deletions as it was done before when using `mutation_summary_gisaid.tsv`?


emmahodcroft:
For line 333 - this was put in because there used to be such long delays (originally) between getting a new `metadata` file (right after `ingest` finished) and getting a new `mutation_summary` file (after the full build) - it was easy to go out of sync (and have a `mutation_summary` file that was older than `metadata`).  So this just prevented me trying to count sequences that had 'no mutations' (when really we just hadn't processed them yet).

Now that all the information is generated in `ingest` this isn't an issue anymore, and we can leave it out. I'm not sure what I meant in the comment about checking mutation & substitution columns aren't empty.. 🤔

emmahodcroft (approved):

PR #275, opened 2022-02-14
implement usecols

MoiraZuber:
Implement pandas read_csv parameter `usecols` to only read selected columns from metadata in order to speed up the script and reduce memory usage.

TODO:

 - [ ] Check list of columns for missing entries (In the line above, `dtype` is provided for columns that are apparently not used throughout the script?)


emmahodcroft:
I've run this on scicore and it runs without errors, so we're capturing all columns. Will merge in!

PR #280, opened 2022-03-07
add aquaria code and altered clusters.json

MoiraZuber:
Added the code provided by Aquaria to `convert_to_web_app_json.py`. It now produces an altered version of `clusters.json` which now contains a field "aquaria_urls", containing a url for each gene. I also uploaded the new clusters.json file.

Possible points of discussion:
- Should "aquaria_urls" be added if no mutations are available (this happens for some joint clusters I think, which are not displayed though). Currently it's added but empty.
-  Make sure I did everything right with the ORF1a& ORF1b vs PP1a&PP1ab naming issue
- Did I correctly decide for the "All mutations per gene" option Andrea suggested? We could also do "Every mutation individually", but I assumed this is not what we want.


ivan-aksamentov:
@MoiraZuber I suggest to avoid constructing the URLs manually as much as possible to avoid broken URLs. URLs have many weird rules and reserved characters, so it often makes sense to rely on libraries, which handle all the edge cases, instead of rolling a custom solution. This will be more reliable and would ensure the best compatibility across browsers.

 - Could you please consider using "url encoding" functionality, either embedded into Python or a 3rdparty library? See for example a guide here: compciv.org/guides/python/how-tos/creating-proper-url-query-strings/. The most convenient is usually to have a dict of key-value pairs which will then be serialized into the kay-value pairs in the URL query string format.
 - For additional clarity and reliability, it would be nice to have a template string in advance and then to substitute the variables into it all at once, rather than concatenating strings as you go.


I was not aware of this feature. Is this something we want to expose in the web UI? I can probably add it fairly quickly, so already know how it will be displayed and see a good location for it, let me know.

ivan-aksamentov:
I added the links to the web app. They are basically the mutation badges with only gene section - same looks and colors.



Is this reasonable?

MoiraZuber:
I tried replacing all string concatenations with f-strings. @ivan-aksamentov thanks for the tutorial links! Please let me know in case it's not yet correct.

Regarding the URL encoding, I just copied the code we were given by the Aquarius people. I'm afraid I can neither really help with the encoding nor the redirects. Maybe this is something we could ask them directly?

emmahodcroft:
This is looking super great - and Ivan - I love the badges! I hadn't even thought of something like this, but it's way way better than a list of text links!

Moira's managed to change the concats to `f`. I haven't had much time yet to look into the URL decoding if not using key-pairs (since their URLs don't use `=`), but would it be worth passing everything we add to the URL through ([from the guide page](http://www.compciv.org/guides/python/how-tos/creating-proper-url-query-strings/#invalid-characters-for-urls)):
```
from urllib.parse import quote
qstr = quote("Stanford University")
```
So that we are covering for any unsupported characters?

>Note that the order of keys is not defined for objects (dicts), so the badges can come in any order. If order is important we should make it an array.

This is a good point actually - I think consistency on webpage is helpful - if people are trying to open up multiple `N` structures from different pages, it's annoying to not have a vague idea of where in the list it appears. My JSON formatting is always sketchy but I think to convert this to array in `clusters.json` would be changing to something like:
```
      "aquaria_urls": [
        {"N": "https://aquaria.app/SARS-CoV-2/N?D3L&R203K&G204R&S235F"},
        {"ORF8": "https://aquaria.app/SARS-CoV-2/ORF8?Q27*&R52I&Y73C"},
        {"S": "https://aquaria.app/SARS-CoV-2/S?H69Del&V70Del&Y144Del&N501Y&A570D&D614G&P681H&T716I&S982A&D1118H"}
      ],
```
Is that right?

If that's correct, and would work with Ivan's code, could you perhaps switch to this Moira? Ideally it would be nice to have `S` first and then the others can just be alphabetical (I imagine S will, as always, get the most attention). But if that's super-fiddly we can just do the whole thing alphabetical by gene. 

Finally,
>I noticed that after navigation to the target URL, browser subsequently performs one or multiple redirects. You can tell by looking to the URL bar. 

Yes, I'm not 100% what's it's doing here - it looks like it's mapping to a particular existing protein structure, but I've no idea how. This would be something we could ask them about when we send a preview, but I imagine this may be some processing happening on their end that may be more difficult to skip around (or hopefully when they sent the URL code they would have done so!)

ivan-aksamentov:
@emmahodcroft 
> My JSON formatting is always sketchy but I think to convert this to array in clusters.json would be changing to something like

That will be difficult to deal with. I suggest:

```json
"aquaria_urls": [
  { "gene": "N", "url": "https://aquaria.app/SARS-CoV-2/N?D3L&R203K&G204R&S235F" },
  { "gene": "ORF8", "url": "https://aquaria.app/SARS-CoV-2/ORF8?Q27*&R52I&Y73C" },
  { "gene": "S", "url": "https://aquaria.app/SARS-CoV-2/S?H69Del&V70Del&Y144Del&N501Y&A570D&D614G&P681H&T716I&S982A&D1118H" }
]
```

> but would it be worth passing everything we add to the URL through 

Yes, passing every component of the ampersand-delimited list is necessary. In the example above, in gene ORF8, there is an asterisk (`*`). This is a reserved character and should be encoded.

The urlencode functions can accept strings, that's their basic form. Sometimes they can also accept arrays instead of dicts, so that the query is produced without `=` signs, or sometimes you just pass a dict with desired keys, but with null values. It's been a while I did this in Python, so don't remember exactly which functions do what.

emmahodcroft:
Ok, this sounds like a plan.

@MoiraZuber Can you plan each element that goes into the URL (ex: `Q27*`, `R52I`) to the command below, so we are sure we're encoding everything right - then put them together into the URL?
```
from urllib.parse import quote
qstr = quote("Stanford University")
```

Can you also investigate if you could put the JSON format into the one Ivan suggests above, so that the order is stable when it appears on webpages? Alphabetical order will work if that's easiest - otherwise `S` first, then the rest alphabetical would be nice, but no deal-breaker if complicated!

After this, I think we'll be almost ready to go!

MoiraZuber:
All components are now passed through` quote()`. `Q27*` now appears as `Q27%2A`. I also adjusted the dictionary format as suggested by @ivan-aksamentov and sorted alphabetically, with S as the first entry.

emmahodcroft:
This looks fantastic Moira - thanks so much for doing this - I know it's fiddly 😆 
But as with the 2-week date issues... turns out better to be careful at the start, and avoid kicking yourself later 😉 

Once we check out the preview, and @ivan-aksamentov can give a 👍 , we can send to Aquarius just to check all's good on their end, and then make this reality!

ivan-aksamentov:
I adjusted web app to the new format

emmahodcroft:
I've checked the URLs with `*` - they seem to work fine! I haven't tested so many phones yet, but the first couple of iPhones, seems to be fine for CoVariants (the aquaria app seems to be a little laggy on mobile, but that's not our problem).

emmahodcroft:
On iPhone 8, OS 13, Safari (selected at random) the defining mutations run over the structure panel, and sometimes the Aquaria protein panel:
image
image

This isn't a problem on iPhone 13, OS 15 Safari, but is on iPhone 12 OS 14...

ivan-aksamentov:
@emmahodcroft let's keep that in mind and say it's a separate issue, not related to this PR

emmahodcroft:
I'll send a link over to Aquarius folks and see what they think!

ivan-aksamentov:
@emmahodcroft Perhaps this what what meant, but note, that the link you've sent in the email is the link to the deployment specific to the commit [3738038](https://github.com/hodcroftlab/covariants/pull/280/commits/37380380969bfeaabad4a1cc964476d48bbb7b11). This is the latest commit on this branch currently.

```
https://covariants-78klv23wl-hodcroftlab.vercel.app/
```
This link will always point only to that particular version.

However, the link in the Vercel's bot first message:

```
https://covariants-git-addaquariaurls-hodcroftlab.vercel.app/
```

is specific to this whole branch and is updated with new deployments on every commits.

Both can be useful, depending on situation.

emmahodcroft:
The Aquaria team noticed that we have an entry for an Aquaria section on the mutation pages, but no links that go to Aquaria. I agree this is a bit odd:
image


Originally I had thought these aren't really worth including for single mutations, but if that's the simpler solution, it might be worth considering, as I think it would just be an extension of @MoiraZuber 's script?

Alternatively we find a way to remove this from showing up if there's no links present for that cluster/mutation.

Apart from the above, I think we are good to go, through Aquaria is asking if we can delay a few weeks while they fix URL stuff on their side. @ivan-aksamentov, I know this is generally not great as it introduces the possibility of merge confusion. If that's a big concern, we could merge in quietly but wait to announce on Twitter. What do you think?

MoiraZuber:
Adding aquaria links for individual mutations should be no problem. However I think we'd need to add some more info on the individual mutation to `clusters.py` then, in the same manner it's found in the `mutations` -> `nonsynonymous` field in the other clusters (including "right" and "left" additionally to "gene" and "pos").

ivan-aksamentov:
I'll make the section to disappear if the array is empty, just in case.

@emmahodcroft We can delay, not a bit deal. It's a small and relatively isolated PR.

emmahodcroft:
>However I think we'd need to add some more info on the individual mutation to clusters.py then, in the same manner it's found in the mutations -> nonsynonymous field in the other clusters (including "right" and "left" additionally to "gene" and "pos")

Good catch Moira, I'll try to add these in now. Shouldn't take too much work.

emmahodcroft:
Ah ha - no, I see the issue here. We should not allow visualization of mutations because I don't track a specific mutation (as in `S:N501Y`) I track _all_ mutations at position 501. So, we couldn't tell Aquarius what to visualize - since the mutation page (and tree) cover any and all mutations at that position.

So in that case, Ivan's fix should have us set!

emmahodcroft (approved):

emmahodcroft:
**This PR is ready to merge pending Aquarius' ok** We'll want to write a tweet and announce (in coordination with them).

PR #282, opened 2022-03-08
feat: add mutation counts data and table

ivan-aksamentov:
Resubmit of https://github.com/hodcroftlab/covariants/pull/219

138282827-20f14d93-2ca2-4497-bd79-d2254dda7e6d


MoiraZuber:
I added Cornelius' script for downloading the mutation count data to `covariants/scripts/` and tidied up the output a bit. @ivan-aksamentov the output should now (hopefully) comply with your expectations. The files can be found in `web/data/mutationCounts/` and are separated into `S` and `nonS`. Please let me know if they still need adjustment somehow.

@emmahodcroft I'll next try to add a new field `snps_with_base` to clusters.py that can be used to apply the script to all clusters (currently it's only for two manually added examples). I assume this is best done manually using covariants.org as a reference?

MoiraZuber:
The script now works for all clusters in clusters.py that are not `type mutation` or `type do-not-display`.

MoiraZuber:
Sorry, yes they're not yet on GitHub. Should I commit them? There is quite a large number of files, so I wasn't sure. They can be created by simply running `python scripts/mutation_counts.py` from within `covariants`.

ivan-aksamentov:
I modified python script to:
 - [x] use less verbose way of iterating over the clusters
 - [x] use `cluster.build_name` as keys for produced JSON entries. Web app does not have access to the top-level keys in clusters.py.
 - [x] merge *S.json and *nonS.json file pairs into single files

I modified web app to:
 - [x] render both tables
 - [x] remove limit on number of rows of these tables

To discuss:
 - Number of rows should be reduced. There are way to many in some clusters.
 - Where table should be situated?


It's been a while and I don't remember anything about this feature. What else needs to be done here?

emmahodcroft:
Thank you @MoiraZuber & @ivan-aksamentov ! I think this is looking great so far. I've asked Cornelius (in slack) to just check the mutations & ensure they look like ones that make sense, since he's pretty familiar with them (sanity check).

_Update_: The long deletions we observe (`ORf1a` and `N`) are not real - we need to address this:
image



## Here's what remains before we'll be good to go 😄 :
Bigger things:
- [ ] Address the deletions at start of ORF1a and end of `N` (these are caused by incomplete sequences & a bug in how nextalign & covSpectrum interact - [as confirmed by Cornelius](https://neherlab.slack.com/archives/C01H7P3TY1M/p1648828737474119))  -- see A below
- [ ] Control for `N` (no coverage) values at positions, which are counted as reference by covSpectrum -- see B below

Smaller things:
- [ ] We should try to create something so that after the main run, we calculate this automatically - @ivan-aksamentov would it be ok to add this to make web-data, or no? I don't want to overload this too much, but the fewer things to remember to run, the better 😆 
- [ ] Decide how many rows to show - I'd say perhaps 10?
  - [ ] @MoiraZuber We should probably apply this in the script, to avoid overloading the app (so only write out the top X muts)
- [ ] Decide where to position list on the page
  - At the moment I'd likely suggest between distribution graph & structure  (After Aquaria link box, when merged) ?
- [ ] Is it possible to change the tables to be 'top-aligned', so that even if S has fewer muts, both tables start at the same place? Otherwise it can be easy to miss that there's an S table

### A 
@MoiraZuber -- to address the first point, Cornelius suggests we add something quite simple to the (every) query, so that we only look for complete sequences: `Orf1a:1., N:420.` (according to Cornelius this should filter to only return sequences that have a value at these positions, meaning that they're 'full sequences'). Could you add this to the query and see if that fixes these long deletions?

### B
Also Moira (sorry!) - Cornelius says that `N` values at a position will be counted as 'reference' by covSpectrum. `N` normally means 'we don't know' - there's no coverage, so it really should be discounted entirely from our calculation of the frequencies. It's hard to know how much this may be impacting our frequencies, but it's possible it could be. One way to control for this (also suggested by Cornelius) would be, for any mutation we want to include in our table (say, top 10), we also query that position, but with `N` as the value: `ORF1a:V1887N` - and subtract this from the denominator ('total') before we calculate the frequencies. What do you think - would this be hard to do? We could also try it manually for a few mutations - but there's always a chance that some parts of the genome are more likely to have `N` than others. Happy to discuss further!

----

Last thought: @ivan-aksamentov It would be cool to allow people to link to the full list somehow, but I don't know if this may just be way complex. For one, we'd have to move the truncation of the list from script to app, which may be overhead, then we'd have to have a link that goes somewhere. So unsure if this is worth pursuing. Any thoughts welcome!

ivan-aksamentov:
@emmahodcroft 

> allow people to link to the full list somehow

I can make a "Show more" toggle between filtered and all rows. I'd need to know which rows to show in filtered mode - either simply top N rows, or having a flag for every row. The disadvantage is that users will download hidden data regardless. We might slit into  2 files - with filtered and with full data. 

We can also make something like a separate page for the full tables if needed.

in any case, when there are many rows it looks weird as you see.

I'd say let's ship it initially with 10-20 rows truncated in Python, and then see how to extend later.

ivan-aksamentov:
There's been a mishap: an unrelated commit https://github.com/hodcroftlab/covariants/commit/de5b58d77952d7d6d19e9bdb576438638b1b45be got into this branch. On Moira's demand, I reverted it in https://github.com/hodcroftlab/covariants/pull/282/commits/07ea71b1138fb11d873283354a9c1f1d98ca4ffc

chaoran-chen:
Hi everyone. About B: It is certainly a big problem for the API itself. I couldn't find time to fix it (in an efficient way) but I think that I will be able to get to it by the end of the month.

MoiraZuber:
I restricted the tables to the top 10 entries.

A: I added the `Orf1a:1., N:420.` filter to the query, which successfully removed the long deletions.

B: I think this should be doable. Should I attempt to implement this now so we can go forward with this, or do we want to wait until the covSPECTRUM side fixes the issue? We could also add a temporary fix now, which can then be removed as soon as the issue is no more.

emmahodcroft:
That's amazing, thank you so much Moira! I really like how this is looking with the truncation and also with those strange deletions being gone :D 

At the moment, if Chaoran thinks he can get to this by the end of the month, I'd be tempted to leave **B** and let covSPECTRUM fix it at source - this would be tidier than trying to do it on our end (most likely). However, @MoiraZuber If you think it's something that could be easily done, we could try it and see what difference it makes to the numbers. I'm on the fence about it, really. (If I had a better sense of how much it might be impacting %s, perhaps I'd feel differently!)

If we get everything else done and this is blocking us, then I'd probably vote to go without it, and update one way or the other later.

Here's an updated to-do list of things, with assignments

- [ ] Add to `make web-data` so that this runs automatically after the main run (@emmahodcroft probably)
- [ ] Move tables to appear below 'distribution graph' (and 'Aquaria link' in future) (@ivan-aksamentov )
- [ ] Change tables to be 'top-aligned' so that even if S has fewer muts, they start at the same horizonal space (@ivan-aksamentov )
- [ ] Add a 'small title' to each table to clarify what they are - "Mutations in S" & "Mutations in Other Genes" or similar (Perhaps just "In Spike" & "Outside of Spike") (@ivan-aksamentov )

chaoran-chen:
Hi. The problem has been solved in LAPIS. Unknowns are now excluded from the mutation proportion calculations.

ivan-aksamentov:
I made the following changes, as discussed above:

 - [x] add captions for mutation counts tables
 - [x] top-align mutation counts tables vertically
 - [x] move mutation counts tables between distribution graph & structure

emmahodcroft:
@chaoran-chen Apologies for the slow reply - I've been on holiday & then working remotely! Fantastic that you managed to get this fixed, this is great!

@ivan-aksamentov I'm loving how the tables are looking now! I think this is brilliant. 

Sounds like the last thing is up to me - add the scripts to the `make` file so this too can run easily as part of updating the data :)

emmahodcroft:
@ivan-aksamentov - I had one more thought - we are pulling this data from CoV-Spectrum, so it would be nice to acknowledge them. I've added this so the title of the card gives credit, but perhaps this should be linkable? (I wasn't sure how to do that) or use a logo?

emmahodcroft:
I've added the mutation-counts script to the makefile, and also added an FAQ explaining the new panel. 

I think we should be about ready to go, if all looks good!

ivan-aksamentov:
@emmahodcroft I could not find a logo.

@chaoran-chen Can you provide your preferred CoV-Spectrum logo which we could add to the table? Which URL would be the best to link to? Or perhaps you could add it yourself, for example inside [CardHeader here](https://github.com/hodcroftlab/covariants/blob/2b4264ccbb974d165a0aed3a4beb03546891b6c1/web/src/components/MutationCounts/MutationCountsSummaryCard.tsx#L113) or inside CardBody just below it?

We have a component for external links: [LinkExternal](https://github.com/hodcroftlab/covariants/blob/2b4264ccbb974d165a0aed3a4beb03546891b6c1/web/src/components/Link/LinkExternal.tsx#L33) (opens links in a new tab and prevents some tracking and security risks.

chaoran-chen:
We don't have a proper logo. Just write the name would be perfect and I suggest linking to `cov-spectrum.org`

emmahodcroft:
I'll try to add a link to the card!

ivan-aksamentov:
@emmahodcroft I addded it. Please double check!

emmahodcroft:
I was trying (and perhaps not 1 million miles in the wrong direction!), but you were much faster @ivan-aksamentov 😆 

image

emmahodcroft:
This looks great! I think we're ready to go, pending one reply from Moira (on slack, to ensure I haven't screwed up the order again), and writing a tweet thread. I'll aim to try and get this up tomorrow!

PR #286, opened 2022-03-08
chore: upgrade web dependencies

ivan-aksamentov:
It's been a while, so why not.

Among huge upgrades are: Next.js 12, React 17, Webpack 5 and Typescript 4.6. But almost every package got upgraded and many got a new major version or even jumped over a few.

The builds should be up to 10 times faster due to removal of Babel and Next.js moving to the new Rust-based Typescript compiler. Deployment will still be slow due to massive amounts of JSON data to be copied, unfortunately.

I feel like the app might have gotten a tiny bit faster as well. But may also be a placebo effect.

Let's see if anything breaks.


ivan-aksamentov:
@emmahodcroft I fixed the Swiss map image.

If there's nothing preventing it, please merge this PR after the next data release, such that we could see this new code running on master for some time. After merge, I will also update the existing PRs with these upgrades. Would be nice to have some time window before the next release to thoroughly check everything.

emmahodcroft:
Thanks Ivan!

Ok, I'll run today's update, then merge this in. I'm currently updating on Tuesday & Thurs, so this will give us until Tues before another push live.

emmahodcroft (approved):

Issue #288, opened 2022-03-14
More descriptive page titles


azizLIGHT:
Many pages have a nondescriptive title 'CoVariants'. 

Pages like Home, Variants, Per country, Per variant, Shared Mutations, and Acknowledgements all have a HTML title of `<title>CoVariants</title>`, with the exception of the FAQ page, which is titled `<title>CoVariants: FAQ</title>`. 

Please consider using more descriptive page titles, like how the FAQ page currently is, for all the other pages, so that browser tabs can have useful titles and browsing can be done easier.

ivan-aksamentov:
This is implemented, but someone needs to take time and fill-in the content here:

https://github.com/hodcroftlab/covariants/blob/2f2f26715e126798ccb2fcac283a137c32b5914e/web/src/seo/seoMetadata.ts#L18-L24

Contributions are very welcome!

azizLIGHT:
That's what I saw as well but wasn't sure because it seems a little 'hard-coded'. 

How can I use the already existing title of the page that exists in the content, and add it to <title>?

For example, /per-country has a in-page <h1 class="PageHeading-sc-1380ypn-0 gStIVs">Overview of Variants in Countries</h1> and that should be the title for this page probably. How can I refer to that title programmatically in <title> inside  seoMetadata.ts. I would like to avoid writing it out manually in seoMetadata.ts, in case the title in <h1> changes, then you have a different and old title inside seoMetadata.ts that will have to be changed again manually.

ivan-aksamentov:
I guess you could extract the existing titles into variables or into a map, and then import into both places.

Alternatively, instead of this centralized solution with a map and `Document` component, one could use the `Head` component on every page instead: https://nextjs.org/docs/api-reference/next/head. This way the SEO things would be in close proximity to the content. I haven't explored this solution, there might be drawbacks as well.

Feel free to play with different approaches and try different things if you have time.

emmahodcroft:
This was another post-Christmas PR that I missed entirely - sorry!
Something that auto-updates would be a nice touch! However in the meantime, I updated these manually.  See #289

PR #289, opened 2022-03-15
Update SEO metadata

emmahodcroft:
As raised in #288 - not all pages have updating titles or descriptions.

This is a quick manual update so that we have some more specific details for each page. (I scripted the part with the variants/mutations.) 

@ivan-aksamentov , are there restrictions on what symbols can be used in these headers/descriptions? I used the `build_name` object from `clusters.py` - so it might have things like spaces and parenthesis. I think this should be fine, but if you could eyeball it quickly, that would be good!

Also, I left out the `image` variable for the other pages as we don't have specific images - is that the right way to do that?

### DIFF
--- a/web/src/seo/seoMetadata.ts
+++ b/web/src/seo/seoMetadata.ts
@@ -41,151 +41,151 @@ export const SEO_OVERRIDES: Record<string, PageMetadata> = {
     name: `${PROJECT_NAME}: Acknowledgements`,
     description: `${PROJECT_NAME}: Acknowledgements`,
   },
-  '/variant/20I.Alpha.V1': {
+  '/variants/20I.Alpha.V1': {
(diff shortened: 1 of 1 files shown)


emmahodcroft:
Hm... the deployment is working for the 'main' pages (Per Variant, Shared Mutations, etc) but not for the variant or mutation pages. I think I see how I messed up.

azizLIGHT:
If merged, this will solve the issue I described and can be closed. 

Maybe automated page titles and descriptions should be a separate issue to be tracked separately.

ivan-aksamentov:
I am not against merging. It's tedious to update this way, but if not updated the missing page will just show the defaults. So not a big deal.

emmahodcroft:
Ok, I'll merge this in now. But better solutions welcome!

PR #290, opened 2022-03-15
fix(web): add missing country flags

ivan-aksamentov:
This adds a few country names to the exceptions list, in order for flags to be rendered correctly.

This is due to inconsistent naming we use, which is not compliant with ISO-3166-1 Alpha-2 Code and with the the flag library we are using.


ivan-aksamentov:
@emmahodcroft 
Just a quick note:

```
  'Saint Martin': 'SX',
  'Sint Maarten': 'SX',
``

are probably the same things. Check on Python side that there aren't any bugs.

emmahodcroft:
Hmmm. Good spotting. We take things straight from the metadata file in Nextstrain, and don't change anything within CoV that I can remember.
@MoiraZuber would you mind checking the `ncov-ingest` info to see if these are sometimes not being corrected?

PR #296, opened 2022-04-13
Hide "also known as" if no alt display name exists

chaoran-chen:
This hides the "also known as" text in the variant title if there is no alternative name.

How it is right now:

image

### DIFF
--- a/web/src/components/Variants/VariantTitle.tsx
+++ b/web/src/components/Variants/VariantTitle.tsx
@@ -25,7 +25,7 @@ export function VariantTitle({ cluster }: VariantTitleProps) {
       <ClusterNameTitle>{cluster?.display_name && `Variant: ${cluster?.display_name}`}</ClusterNameTitle>
       {
         <ClusterNameSubtitle>
-          {cluster?.alt_display_name && (
+          {cluster?.alt_display_name?.length > 0 && (
             <>
               {`also known as `}
               {cluster.alt_display_name.join(', ')}


emmahodcroft:
Thanks for spotting this @chaoran-chen ! I hadn't noticed this little bug. 🙈  @ivan-aksamentov if this looks good to you, I'm happy to merge.

ivan-aksamentov:
@emmahodcroft 
Looks okay. But the main reason for it to happen is the empty strings for alt names in the data. They should not be there, I think. Feel free to either:
 - merge this
 - or to remove empty strings from data
 - or both

chaoran-chen:
Hi @ivan-aksamentov. I only looked very briefly at the code, so I might be talking nonsense; but from what I am seeing, there are no empty strings. `alt_display_name` is defined as `string[]` ([code](https://github.com/hodcroftlab/covariants/blob/ac1c6cb009bf3e7f92b383bcecb8640a6e999105/web/src/io/getClusters.ts#L19)) and in case of e.g. 21I, it's just an empty array ([code](https://github.com/hodcroftlab/covariants/blob/da3868c6bc1a0a8ecc037f98cd69288aebbf1d27/scripts/clusters.py#L241)).

ivan-aksamentov:
@chaoran-chen  Ah, sorry, I confused something. I meant empty arrays, not empty strings.

The Typescript definition is 
```ts
alt_display_name?: string[]
```
which means "missing or array of strings".

So then I wrote the check in the component

```ts
cluster?.alt_display_name
```

I assumed that lack of alt name will be expressed as a missing field, but in the data it seems to be expressed as empty arrays.

emmahodcroft:
Ah ha - I see what you mean Ivan. I would be happy to do 3 - both. I think it's nice if we can handle things like accidental empty arrays neatly (not unimaginable it happens again), but still do this kind of tidy-up. 

I'll try to make this change to tidy up these 'empties' in `clusters.py` on master.

emmahodcroft:
I'm just letting that build now to check that I didn't manage to break anything, but if that looks good, and Chaoran's changes look good, feel free to merge this Ivan!

ivan-aksamentov:
@emmahodcroft 

> removing these empty arrays from clusters.py didn't seem to solve the underlying problem on master

Web data needs to be updated to regenerate `web/data/clusters.json`

I am merging this PR, because it's seems to be low risk.

ivan-aksamentov:
@chaoran-chen Could you please add yourself to the list of contributors as explained here:  https://github.com/hodcroftlab/covariants/pull/48

emmahodcroft:
Ah, of course, sorry Ivan for the oversight - and thank you!

PR #298, opened 2022-04-22
add category unknown to case count data

MoiraZuber:
Adding a new category, currently called `Unknown`, to cases data which contains all remaining case counts that are not covered by known clades.


ivan-aksamentov:
I pushed a couple of commits:
 - [x] renamed "Unknown" to "others", for consistency with "Per Country" plots
 - [x] added "others" as an additional area segment on "Cases" plots

Please check that it makes scientific sense.

emmahodcroft:
And yes, good call @ivan-aksamentov to say 'others' for consistency with the other plots (and some of them may have Pango lineages, so they may not be 'unknown' as such). I think this looks fantastic!

PR #300, opened 2022-04-23
fix: ensure url parameters correctly apply for 'Cases' page

### LINKED ISSUE #299: URLs not working on 'Cases' page
URLs on Per Country and Per Variant pages allow the URL to modify if users specify particular countries/variants. This should work the same on the Cases page. Indeed, selecting countries does change the URL at the top (and view). However, when putting that URL in a new window, it doesn't actually filter to just showing what's in the URL.

Compare URLS:

On Cases page:
https://covariants.org/cases?country=USA&country=Canada&country=Japan&country=Croatia Should show:
image
(but shows all countries)

But on Per Country Page:
https://covariants.org/per-country?country=USA&country=Canada&country=Japan&country=Croatia
Works as expected and shows:
image

It also doesn't work on Cases page for variants:
https://covariants.org/cases?variant=21A+%28Delta%29&variant=21I+%28Delta%29&variant=21J+%28Delta%29

I'm hopeful this is just a plumbing issue - the 'Cases' page just wasn't fully hooked up into the URL system, and can be without too much work!

ivan-aksamentov:
Resolves #299 

The URL parameters were not applied on app initialization to the recoil atoms which store the state for countries, continents and clusters specific to 'Cases' page.

This PR adds the logic that applies incoming parameters to the recoil state.

### DIFF
--- a/web/src/components/Cases/CasesPage.tsx
+++ b/web/src/components/Cases/CasesPage.tsx
@@ -8,10 +8,10 @@ import { Layout } from 'src/components/Layout/Layout'
 import { MainFlex, SidebarFlex, WrapperFlex } from 'src/components/Common/PlotLayout'
 
 import { getPerCountryCasesData, filterClusters, filterCountries } from 'src/io/getPerCountryCasesData'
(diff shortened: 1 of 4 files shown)


emmahodcroft:
This is working great now Ivan, thank you!

PR #306, opened 2022-04-29
fix: avoid duplicate keys for different recoil atoms

ivan-aksamentov:
This fixes a copy-paste mistake, where 2 Recoil atoms ended up having the same key. I renamed the key for one of the atoms.

Duplicate keys could have caused app to misbehave or crash.


ivan-aksamentov:
@emmahodcroft This one is an important bugfix for a bug that might have caused instability.

emmahodcroft:
I'll go ahead and merge this in then!

PR #308, opened 2022-05-03
Add 22A, B, C

emmahodcroft:
Making a PR to test full-blown trial of adding "the 22s" - pages, plotting, the whole shebang.


ivan-aksamentov:
It needed a newline between the import and content. It could not tell where one ends and another starts.

ivan-aksamentov:
How do I know? I had no idea, but I renamed the file to .mdx, and then WebStorm told me.

At this point it probably does not make sense to keep the .md extension, because these are not really Markdown files anymore, but mostly MDX. Perhaps if we rename, it will be easier for you to find buds using your text editor's capabilities?

Issue #315, opened 2022-06-14
Sort countries alphabetically on cases page


yuval-harpaz:
Please sort the charts of countries' cases by alphabetical order, or make a "sort" button to do that. It takes time to browse for my country, I end up using ctrl+f. Thanks
[This](https://covariants.org/cases) page

dcordoba97:
Hello @yuval-harpaz ! Is anyone working on this issue? If not, I'd love to :)

emmahodcroft:
Hi @dcordoba97 - I don't think anyone is actively working on this, but there was a [previous PR](https://github.com/hodcroftlab/covariants/pull/317) that would be worth looking over!

One thing worth noting is that we'd need a button or slider somewhere on relevant pages (Per Country, Per Variant, Cases) to allow sorting by either number of sequences (default) or alphabetical (on choice), so that users can select which they prefer! (Rather than just switching all sorting to alphabetical, as the other PR did.)

With a note to [the issue about the fact](https://github.com/hodcroftlab/covariants/pull/317#issuecomment-1156294151) that the sorting on the Cases page didn't seem to be strictly by sequence counts - it would be fine to implement something that makes this work as intended, rather than however it currently is!

dcordoba97:
Sounds interesting! Thanks for the guidance!

PR #317, opened 2022-06-15
sort coutries alphabetically on case page

umaxyon:
issue #315 

image

### DIFF
--- a/web/src/components/Cases/CasesPage.tsx
+++ b/web/src/components/Cases/CasesPage.tsx
@@ -38,6 +38,7 @@ export function CasesPage() {
     const { withCountriesFiltered } = filterCountries(countries, perCountryCasesDistributions)
     const filteredClusters = filterClusters(clusters, withCountriesFiltered)
     const { enabledClusters, withClustersFiltered } = filteredClusters
+    withClustersFiltered.sort((a, b) => (a.country > b.country ? 1 : -1))
     return { enabledClusters, withClustersFiltered }
   }, [countries, perCountryCasesDistributions, clusters])
 


emmahodcroft:
Hi @umaxyon - thanks for this! However, the current way that countries are sorted is based on the number of sequences, which provides a nice top-to-bottom gradient of (generally) how confident we can be in the patterns we see. I would be happy to have an option to change this sorting to alphabetical, but it should be a button or slider on the web-page that allows this view to change depending on what the user wants to see (either by sequence count or alphabetically). 
I'd be happy to integrate something like this, if that's something you're able to do!

umaxyon:
Hi @emmahodcroft 😃 
I gave up because I did not know the original sort order criteria.
It is neither the number of "distribution" arrays nor the sum of "total_sequences", but a subtle difference.

```
USA            :	 distribution size: 55:	sum totla_sequence: 3168562
United Kingdom :	 distribution size: 55:	sum totla_sequence: 2637606
Germany        :	 distribution size: 54:	sum totla_sequence: 632128
Denmark        :	 distribution size: 55:	sum totla_sequence: 509391
France         :	 distribution size: 55:	sum totla_sequence: 285611
Canada         :	 distribution size: 55:	sum totla_sequence: 316619  <-- ? 
Japan          :	 distribution size: 55:	sum totla_sequence: 298010
Sweden         :	 distribution size: 54:	sum totla_sequence: 186642
Switzerland    :	 distribution size: 55:	sum totla_sequence: 129169
...
```
I used the following debugging code
```javascript
// CasesPage.tsx  line 41
  withClustersFiltered.forEach(d => {
    const total = d.distribution.map(d => d.total_sequences).reduce((a, b) => a + b, 0);
    console.log(`${d.country.padEnd(15, ' ')}:\t distribution size: ${d.distribution.length}:\tsum totla_sequence: ${total}`)
  });
```
sorry. good luck!

emmahodcroft:
Hmm, that's interesting. I'll have to double-check how this is calculated. In fairness, I think I may 'steal' the order from the Per Country page, and it's possible that if we don't have case data to match the same time periods, some sequences are 'excluded' and thus the counts don't technically match. But I'll try to look into this!

PR #319, opened 2022-06-16
perf: use another, faster charts library - echarts

ivan-aksamentov:
Reimplements charts using [echarts](https://github.com/apache/echarts) instead of [recharts](https://github.com/recharts/recharts).

eharts is based on HTML5 canvas (as opposed to SVG in recharts), and should render much faster.

As a bonus, echarts supports pan&zoom and many other features.


emmahodcroft:
Hi @ivan-aksamentov I'm sorry for such a delay here! I've just tried this out for Per Country and it's AMAZING. It's so smooth, looks so good, and I love that the overlay boxes are shorter! I ALSO LOVE THE ZOOMING!!! 

I actually think this is almost great as it is, but we would need to add counts to the mouse-over-box (as currently shows on Per Country).

Optionally it would be super cool to know how possible the following would be:
- Potentially get rid of the Y-axis line thing - I'm just not sure it's informative. But also no strong feelings. Maybe it is useful?
- Can we set it so that to zoom, you press CTRL+SCROLL rather than just SCROLL? I found the zoom by accident while trying to scroll down the page. I love it - but - it would be good if it doesn't happen when you're trying to scroll the page!
- Can we add back in some kind of grid bars to be plotted, as we currently have in light grey? I imagine with such a powerful plotting program this may be possible and even be able to scale with zooming. I just think they're a nice touch.

I'm super excited by this! I assume we could also extend to Per Variant & Cases?

emmahodcroft:
Oh - there is also one small issue - the colors don't seem to be matching up in the mouse-over 🙃 

image

(Actual colors:
image

)

ivan-aksamentov:
- [ ] add counts to the mouse-over-box (as currently shows on Per Country)
 - [x] get rid of the Y-axis line thing for the mouseover
 - [x] zoom only when Shift key is pressed. Ctrl cannot be used because it triggers built-in browser's page zoom (at least not easily).
 - [x] add grid bars 
 - [x] Also port Cases plots
 - [ ] Also port Per Variant plots

richardgoater:
Screenshot 2022-07-23 14 50 33
Hi folks! Just checking this out and I think the dots are still different colours? I could be wrong.

emmahodcroft:
Good catch @richardgoater - it looks like at least the mouseover legend matches the plot now, but you're right that the dots aren't matching. 
For me this is a little less critical - I'd be ok if they were just black/grey/neutral colour (if that's easier) - or if they matched. But having them coloured but not matching is a bit odd!

emmahodcroft:
This makes me sad because of the zooming. :(

richardgoater:
I will implement zooming soon! 🙏🏽

Issue #323, opened 2022-07-04
Optional "Three letter" Amino-Acid badges for beginners


chatelao:
IS
- Mutations are shown with the one-letter AA abbreviation.
- As a "amateur" bio chemistry nerd, I'm not very fluent with this code (yet)
- The TLA (Three letter acronyms) are much more understandable 

COULD
- Having a switch between "pro (one letter)" and "amateur (three letter) muatation badges would be very nice

https://github.com/hodcroftlab/covariants/blob/b253cd94e4dcf4f32b3f7667717304d086c16141/web/src/components/Common/MutationBadge.tsx

emmahodcroft:
Hi @chatelao - Thanks for using CoVariants and thank you for this suggestion! I agree, this would be a very cool addition. Unfortunately at the moment we don't have the resources to implement this, as our project is pretty small. However, if someone else would like to work on this, I'd welcome PRs to add this feature!

carlos-montano-hub:
Hi, I'm interested in this issue and project, I'll take a look at it and let you know of my advances

emmahodcroft:
Thanks @carlos-montano-hub - that would be great! It would be fantastic to have some kind of toggle in the header or footer of pages that could switch this on/off across the site :)

chatelao:
Whenever it helps @carlos-montano-hub - Here the full list as "csv"

```
"Abbreviation (1 Letter)" ,"Full Name"     ,"Abbreviation (3 Letter)"
"F"                       ,"Phenylalanine" ,"Phe"
"L"                       ,"Leucine"       ,"Leu"
"S"                       ,"Serine"        ,"Ser"
"Y"                       ,"Tyrosine"      ,"Tyr"
"C"                       ,"Cysteine"      ,"Cys"
"W"                       ,"Tryptophan"    ,"Trp"
"P"                       ,"Proline"       ,"Pro"
"H"                       ,"Histidine"     ,"His"
"Q"                       ,"Glutamine"     ,"Gln"
"R"                       ,"Arginine"      ,"Arg"
"I"                       ,"Isoleucine"    ,"Ile"
"M"                       ,"Methionine"    ,"Met"
"T"                       ,"Threonine"     ,"Thr"
"N"                       ,"Asparagine"    ,"Asn"
"K"                       ,"Lysine"        ,"Lys"
"V"                       ,"Valine"        ,"Val"
"A"                       ,"Alanine"       ,"Ala"
[...]

PR #325, opened 2022-07-05
Add three letter lookup-table for Amino Acids

chatelao:


carlos-montano-hub (approved):
Thanks!
It will be required for #323

PR #330, opened 2022-07-12
Optional Three Letter Abbreviation for amino acid right table

carlos-montano-hub:
Added a button to change the right table from One Letter Abbreviation to Three Letter Abbreviation in order to solve #323

image 
image

In case of this PR being accepted, please accept #325 first, as I copied the code from there and in that way @chatelao  would not lose his contribution 

Any feedback please let me know


web/src/colors.ts
```diff
@@ -92,4 +93,25 @@ export const AMINOACID_COLORS: Record<string, string> = {
...
+  'Lys': '#CEC0F3',
+  'Met': '#C3ED3C',
+  'Asn': '#F29290',
+  'Pro': '#D2D1F8',
+  'Gln': '#F8C4E3',
+  'Arg': '#A6ACEF',
+  'Ser': '#D8B9D4',
+  'Thr': '#F0D6E3',
+  'Trp': '#86B0CC',
+  'Tyr': '#8FC7D1',
```
ivan-aksamentov, line 116: Any reason for duplicating colors? Duplication is typically bad and increases maintenance and probability of mistakes.

ivan-aksamentov (changes requested):
If the goal is to modify how amino acids letters are rendered, is it not logical to change the actual rendering code, rather then patching the source data somewhere half-way?

Here is where the letters are rendered: 
https://github.com/hodcroftlab/covariants/blob/a3ef0c79da438947ebdae4dcf118d6a41c764222/web/src/components/Common/MutationBadge.tsx#L168-L170

ivan-aksamentov:
I was hoping that we could implement a global toggle for all badges. Was this not the intent?

Perhaps the side bar on these particular pages is not the best place for a toggle, because when switching to other pages user loses access to the toggle.

ivan-aksamentov:
> In case of this PR being accepted, please accept https://github.com/hodcroftlab/covariants/pull/325 first, as I copied the code from there and in that way @chatelao would not lose his contribution

You could have based your branch on chatelao's branch. This would simplify things a lot. It's not too late to rebase.

carlos-montano-hub:
Hello, thanks for the feedback @ivan-aksamentov @emmahodcroft  I will be working on it and let you know

web/src/colors.ts
carlos-montano-hub: I think we have two options for the definition of the left/right colors this is one and the other is to translate from TLA to OLA in the color function.

However, I am looking at your feedback and it may not be necessary to change the color component

PR #331, opened 2022-07-15
feat: use Nextclade_pango lineage before snps

corneliusroemer:
First priority remains Nextstrain clade
Second priority becomes Nextstrain_pango
Third priority is SNPs

One thing to bear in mind: this only works if Nextclade pango exists already
Code may want to be refactored to allow for a special Nextclade_pango=true field


emmahodcroft:
Going to merge this into the 22D branch as a trial - thanks @corneliusroemer !

emmahodcroft:
This seems to work!

PR #337, opened 2022-07-21
feat: limit number of initially displayed plots to 12 (echarts version)

ivan-aksamentov:


emmahodcroft:
I guess after Richard G's fix, we should delete these and the branches? But I don't want to it straight off in case there's stuff worth keeping somewhere!

Issue #316, opened 2022-06-14
equal Y axis for cases by country charts


yuval-harpaz:
Please add a field for setting a high limit for Y-axis to all the charts in the [cases](https://covariants.org/cases) page. It would make a fairer comparison between countries and allow zooming in when looking at a new variant. Thanks.
![image](https://user-images.githubusercontent.com/445416/173500584-2d7c73d0-592a-4122-bef8-7d4ef08d1e1d.png)

emmahodcroft:
Interactive axes would be a great addition to CoVariants. Unfortunately we haven't had time to look into this, but we are looking to perhaps switch to a new charting library soon. This may open up more options! In the meantime, happy to take PRs that may improve this functionality!

yuval-harpaz:
Unfortunately I've got zero experience with Recharts, so I cannot add a user input box. But it may be simple to make "other" also disappear when "deselect all". This seems to be what's keeping Y up there

emmahodcroft:
Ah, that's a good point @yuval-harpaz . @ivan-aksamentov would it be possible to add 'other' to the left-hand selection bar on the 'Cases' page, so it can be deselected - and if someone wants, not selected back in again?

ivan-aksamentov:
@emmahodcroft It's not easy to do in the app. There is no "other" variant. Instead, it is currently computed ad-hoc as a difference between the sum of all variants and total - computed for every user on each their visit for every country being displayed, for every time point. And, as any computation done in the covariants frontend, it's a very silly thing to do, because it's constant data.

If you want "other" to behave as a variant, then the right thing to do is to add it as a variant named "other" to the data in the backend scripts. And it will appear automagically. We could then remove the ad-hoc computation in javascript and the corresponding grey areas on the plot.

This should hopefully allow Y axis to rescale.

yuval-harpaz:
Unless you need "other" separately for other uses, I'd consider making a shortcut, leave it there, just compute max for Yaxis ignoring it.

emmahodcroft:
Hi @ivan-aksamentov - that makes sense to me. I'll see if we can work that in!

PR #342, opened 2022-08-09
feat: add "others" as a variant

ivan-aksamentov:
Resolves https://github.com/hodcroftlab/covariants/issues/316

Treating "others" as a variant allows:
 -  to remove unnecessary computation from web app - the data for "others" is immutable, should be precomputed in advance and be stored along with other variant data
 - to have a checkbox which allows toggling it on the plots

### DIFF
--- a/scripts/clusters.py
+++ b/scripts/clusters.py
@@ -1030,7 +1030,7 @@ clusters = {
     },
 
     "22D": {  #BA.2.75
-       "snps": [3927, 26275, 3796], #TGT = C3927T, A26275G, C3796T 
+       "snps": [3927, 26275, 3796], #TGT = C3927T, A26275G, C3796T
        "snps_with_base": ["3927T", "26275G", "3796T"],
        "cluster_data": [],
        "nextstrain_build": True,
@@ -1362,7 +1362,7 @@ clusters = {
         },
     },
 
-    # C.37 
+    # C.37
     "21GLambda": {
         "snps": [7424, 23031, 24138],  # ORF1a 2387 (7424G)  #S 490 (23031C) 859 (24138A)
         "snps_with_base": ["7424G", "23031C", "24138A"],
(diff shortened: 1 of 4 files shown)


scripts/clusters.py
```diff
@@ -2509,4 +2509,16 @@
...
+        "build_name": "others",
+        "cluster_data": [],
+        "col": "#222222",
+        "snps": [],
+        "display_name": "others",
+        "graphing": False,
+        "important": False,
+        "nextstrain_build": False,
+        "type": "variant"
+    }
```
ivan-aksamentov, line 2523: This is the definition of the "others" variant in `clusters.py`. This object will be emitted into `clusters.json` alog with other variants. It should be the last object in the list.

web/data/perCountryDataCaseCounts.json
```diff
@@ -30,7 +30,8 @@
         "22C (Omicron)",
         "22D (Omicron)",
         "S:677H.Robin1",
-        "S:677P.Pelican"
+        "S:677P.Pelican",
+        "others"
       ],
```
ivan-aksamentov, line 35: "others" needs to be added into the list of clusters in json data. Here I illustrate what the result should look like for cases data. This is a manual addition and is not enforced in code. Python scripts will need to be changed for that.

web/src/components/Cases/CasesPlot.tsx
```diff
@@ -90,16 +90,6 @@ export function CasesPlotComponent({ cluster_names, distribution }: CasesPlotPro
...
-                    type="monotone"
-                    dataKey={CLUSTER_NAME_OTHERS}
-                    stackId="1"
-                    stroke="none"
-                    fill={theme.clusters.color.others}
-                    fillOpacity={1}
-                    isAnimationActive={false}
-                  />
-
                   <CartesianGrid stroke={theme.plot.cartesianGrid.stroke} />
```
ivan-aksamentov, line 93: Here I remove the additional ad-hoc area segment from the cases plot. The same should be done for the other plot(s), when they get their own "others" data.

emmahodcroft:
Thanks Ivan! This is working really well from my playing around. I will try to get to editing the code so that these additions are permanent so we can merge this in

PR #349, opened 2022-10-06
Add early check for meta-clusters

MoiraZuber:
Adding a check early in the script that tests whether all components required by a "meta_cluster" are selected in cluster_to_run.


emmahodcroft:
I think this looks great Moira, thank you so much for implementing it, and so quickly! This should avoid asking for something and getting nothing, or wasting time only to find out you asked for nothing 😆 I liked the nice touch of a warning if it doesn't find one of the components, too!

emmahodcroft (approved):

PR #353, opened 2022-10-20
add check for graphing

MoiraZuber:
Add a check for the "graphing" flag of detected daughter clades. If graphing == False, do not add these sequences to any covariants output, however do add them to the nextstrain files used for builds.


emmahodcroft:
Pushing up data to test

emmahodcroft:
It _did_ print out 22E nextstrain file, with the right number of columns :)

PR #356, opened 2022-10-24
22E edits

emmahodcroft:
Minor edits on top of Cornelius' additions


emmahodcroft:
Argh, I wanted this to preview so I could just check all the links work (sometime the markdown falls down and you have to use HTML instead) but then forgot - it's merged now

Issue #311, opened 2022-05-08
Localization of this website


tonyqus:
I'd like to check if you accepts translation contribution of this website. I'm from Shanghai, China. I think the website is really useful for Chinese people since we are still in serious pandamic season.

emmahodcroft:
Hi @tonyqus ! Thank you for reaching out and for your offer. At the moment we don't have any way to support translations, but I'm chatting with the team about whether this is something we could support. I'll try to get back to you! It would certainly be great to make CoVariants more accessible to more people!

emmahodcroft:
At the moment I'm not sure we have the support available that we could make this happen - unfortunately it would require splitting out all the text areas on the website into separate files & then having these pulled in based on the language. While it would be great to support in theory, I don't think this is something our tiny team (3 people, part time) could support right now.

However, one can always translate the website in Chrome or using Google Translate!

tonyqus:
I totally understand that you are a small team and you don't have time to support the localization. 

I analyze your code and figure out that you are using next.js, which can support i18n (localization) easily. And I did some experiment and make it work on my local with some urls like localhost:3000/zh-CN/faq to support Chinese. 

The current situation in China is a bit complex. The government starts giving up zero-covid policy and a lot of people here still believe Omicron is very dangerous due to the previous news from Chinese media and some fake medical experts. They do need a website like what you did to understand the fact of Omicron. Our official cdc website doesn't provide this kind of information. 

The question is do you accept PR so that I can help you make localizaiton for Chinese happen?

tonyqus:
The menu is localized in Chinese
![image](https://user-images.githubusercontent.com/772561/207322208-d43f89bb-3d3b-486b-9f75-dc21a7cdd235.png)

The major translation content is located in content folder. I can successfully load zh-CN.md file when the url routing changes to /zh-CN/xxx
![image](https://user-images.githubusercontent.com/772561/207322612-daeb6950-e79b-456d-b4e8-9bcb999eddac.png)

ivan-aksamentov:
Hi @tonyqus, 

It's not hard to add the technical implementation. Covariants is currently undergoing a rewrite and I've already added `i18next` and `react-i18next` to the new version. But it will take a little while before all other issues are resolved there.

We wouldn't use Next.js's i18n for that. We currently use SSG (`next export`) to make deployment simpler and cheaper, but Next.js i18n requires a server, so that would not work. Neither it is necessary - the same can be done client-side.

[...]

PR #364, opened 2022-12-15
Add localization (internationalization)

ivan-aksamentov:
Resolves: https://github.com/hodcroftlab/covariants/discussions/292 https://github.com/hodcroftlab/covariants/issues/311

This adds internationalization (i18n). Simply speaking the text in the application is now translated to multiple languages, and you can now switch between these languages.

### What's new:
 - [x] there is a dropdown on nav bar, which allows to switch locale (language)
 - [x] the last selected locale is remembered in Local storage of the browser, so page refreshes and repeated visits preserve the locale
 - [x] ~the `lang` URL parameter now allows to set language in the URL, so that user navigating to this URL will see the app in this language (overriding Local storage)~ [it was braking the router; temporarily disabled]

#### Hardcoded strings

Strings hardcoded in the app code (like text of buttons or of nav bar links) are now translated using `react-i18next` ([1](https://www.i18next.com/), [2](https://react.i18next.com/)). Also dates and numbers are translated using other handy packages. All translatable strings need to be surrounded by `t()` function, which comes from `useTranslationSafe()` hook. 

The usual string interpolation in these strings does not work (interpolation will happen before the string is extracted for translation)! So it needs to be replaced with string interpolaton of i18next format:

```js
"Hello, ${name}!" -> t("Hello, {{name}}!", { name: "Covariants" })
```

i.e. interpolated variables should be additionally set using a dict, passed as the second param to the t-function. Try to pick variable names that are different from keys in the dict (i.e. don't use JS shorthand syntax), because any refactor that leads to variable rename will break the interpolation (the variable in the template string will not be renamed by the refactoring tools), and the app will be showing uninterpolated content without any warning.

#### Maintenance procedure for hardcoded strings:
 
 - `yarn i18n:extract`: Extract English strings. The tool parses the app code and takes strings surrounded by `t()` and puts them into [`web/src/i18n/resources/en/common.json`](https://github.com/hodcroftlab/covariants/blob/6ba6e6ce6aa7c1f9b1871ba4cc5857e5107a4c9f/web/src/i18n/resources/en/common.json). 
 - `yarn i18n:extract`: This adds additional words from [`web/src/i18n/additional_keys.json`](https://github.com/hodcroftlab/covariants/blob/6ba6e6ce6aa7c1f9b1871ba4cc5857e5107a4c9f/web/src/i18n/additional_keys.json) to the `web/src/i18n/resources/en/common.json`. These are some of the words which appear in the UI, but are not in the code, for example because they come from data (e.g. names of countries and of continents).
 - `yarn i18n:translate`: This translates the English strings from `web/src/i18n/resources/en/common.json` and creates or updates [`web/src/i18n/resources/<locale>/common.json`](https://github.com/hodcroftlab/covariants/tree/6ba6e6ce6aa7c1f9b1871ba4cc5857e5107a4c9f/web/src/i18n/resources) for each locale we support. This uses AWS Translate paid service and requires AWS credentials configured. But of course the translation can also be done manually, without using any services. For that, just add the necessary key-value pairs to the JSON file corresponding to a language. 
 - `yarn i18n:fix`: Massages the resulting translations, and applies some of the obvious fixes, because AWS Translate is not perfect. For examples, it omits spaces around string interpolation.

#### Markdown content

Markdown content files have been moved from `content/` to `web/src/content/${locale}`. For now there is only `en/` directory for English locale. Other locales should be just copies of the `en/` directory, but with content in the corresponding language.

The names of directories should be the same as [here](https://github.com/hodcroftlab/covariants/tree/6ba6e6ce6aa7c1f9b1871ba4cc5857e5107a4c9f/web/src/i18n/resources), or [here](https://github.com/hodcroftlab/covariants/blob/6ba6e6ce6aa7c1f9b1871ba4cc5857e5107a4c9f/web/src/i18n/i18n.ts#L65-L86), or as the letter code in the language selection dropdown in the app.

Markdown content now should only be loaded using `<MdxContent filepath="${filepath}"/>`, where `filepath` is the path relative to `web/src/content/en` (or any other locale than `en`, they are mirrors).

If localized md file is missing, the English version will be loaded as a fallback.

#### Maintenance procedure for markdown content:

I haven't figured how to automatically bulk-translate markdown yet. It's MDX actually, because we use React tags there. So this will likely need to be done manually. Just keep md files in `web/src/content/${locale}` for each `locale` in sync with English locale in `web/src/content/en`. "Just". It's a large upfront work, and it also requires expert knowledge every time something changes in English version. Typically you need many experts, per language, because rare human knows more than a handful of these.

### Further work:

 - What languages do we need?
 - The app and its code needs to be inspected and the missing hardcoded strings need to be surrounded with `t()`, i.e. "Hello" -> "t("Hello")".
 - check md files included into md files: I tried to replace all imports with `<MdxContent/>`, but might have missed something.
 - check if any of the strings in the app contain extra `$`, `{` or `}` . This might have been left over by me when converting string interpolation from JS to i18next format.
 - write translations for md content
 - obviously, improve translation quality. AWS Translate often generates very poor quality translations.
 - Support [plurals](https://www.i18next.com/translation-function/plurals), cases, declinations, etc. This is a big work - each string needs to be annotated with grammar context. Probably out of our reach.


emmahodcroft:
List of files that contain large chunks of text that would need to be translated if translation were made possible:

covariants/content/:
[Faq.md](https://github.com/hodcroftlab/covariants/blob/master/content/Faq.md)
[Home.md](https://github.com/hodcroftlab/covariants/blob/master/content/Home.md)
[NameTable.md](https://github.com/hodcroftlab/covariants/blob/master/content/NameTable.md)
[PerAreaIntro.md](https://github.com/hodcroftlab/covariants/blob/master/content/PerAreaIntro.md)
[PerClusterIntro.md](https://github.com/hodcroftlab/covariants/blob/master/content/PerClusterIntro.md)
[PerCountryCasesIntro.md](https://github.com/hodcroftlab/covariants/blob/master/content/PerCountryCasesIntro.md)
[RegionBreakdownInfo.md](https://github.com/hodcroftlab/covariants/blob/master/content/RegionBreakdownInfo.md)
[SharedMutations.md](https://github.com/hodcroftlab/covariants/blob/master/content/SharedMutations.md)
[VariantsPageIntro.md](https://github.com/hodcroftlab/covariants/blob/master/content/VariantsPageIntro.md)

Every variant page in the [clusters folder](https://github.com/hodcroftlab/covariants/tree/master/content/clusters)

Each of the country intros in [PerCountryIntro folder](https://github.com/hodcroftlab/covariants/tree/master/content/PerCountryIntro)


I think this is all the pages...?

PR #369, opened 2023-01-12
fix: hide xmas toggle when outside date range

ivan-aksamentov:


emmahodcroft:
Thanks Ivan! I would love to have Christmas all year but I suppose others are happy to keep it less often. This will help!

PR #396, opened 2024-11-27
395 update packages

AdvancedCodingMonkey:


fengelniederhammer (commented):
Nice work! I just have a bunch of question.

web/public/data/params.json
```diff
@@ -1,4 +1,4 @@
 {
   "max_date": "2024-09-23",
   "min_date": "2020-04-27"
-}
\ No newline at end of file
+}
```
fengelniederhammer, line 4: Let's revert this? Posix systems kinda expect a tailing newline
 https://www.baeldung.com/linux/files-end-with-newlines

web/src/hooks/useAxiosQuery.ts
```diff
@@ -76,7 +79,8 @@ export function useAxiosQuery<TData = unknown>(url: string, options?: UseAxiosQu
   }, [res.data, url])
 }
 
-/** Make multiple cached fetches in parallel (and uses `Suspense`, by contrast to the default `useQueries()`)  */
+/** Make multiple cached fetches in parallel */
+// TODO: write a test this function, as it is currently unused
```
fengelniederhammer, line 83:
```suggestion
// TODO: write a test for this function, as it is currently unused
```

Why do we need the test and the function when it's actually unused?

web/next.config.ts
```diff
@@ -93,23 +96,34 @@ const nextConfig: NextConfig = {
...
+  sassOptions: {
+    includePaths: ['node_modules'], // the correct option should be loadPaths but somehow that does not work (yet)
+    // TODO: remove this silencing once bootstrap has applied SASS API changes (https://github.com/hodcroftlab/covariants/issues/402)
+    // There are a lot of warnings coming from SASS API changes that bootstrap has not implemented, silencing those
+    // However, this might mask warnings from our own code, so be sure to check from time to time. `quietDeps` seems to
+    // not work completely as intended.
+    quietDeps: true,
+    silenceDeprecations: ['mixed-decls', 'color-functions', 'legacy-js-api', 'global-builtin', 'import'],
+  },
+  output: 'export', // TODO: with the cli command, two threads were used here, no idea how to do it with this config option.
```
fengelniederhammer, line 108: What about this TODO? Is there an issue that we can reference? Or can we simply delete the comment?

fengelniederhammer (changes requested):
Actually looks good :+1: 

I found one thing - Something is odd with the translation here:
grafik
https://covariants-git-395-update-packages-hodcroftlab.vercel.app/variants/24B.Omicron

fengelniederhammer (approved):
LGTM

Issue #401, opened 2024-11-29
Realign deployment pipeline


AdvancedCodingMonkey:
We have multiple entities involved in deployment (GitHub, Vercel, CircleCI). Streamline the process and add steps like testing, linting, formatting where appropriate.

PR #434, opened 2024-12-03
412 set up testing infrastructure

AdvancedCodingMonkey:
This PR is setting up testing infrastructure for local development. Setting up pipelines is postponed to #401 . I chose to switch to vitest for unit tests as the configuration setup seems orders of magnitude easier than for jest.


web/package.json
```diff
@@ -249,9 +249,9 @@
     "i18next-parser": "^9.0.2",
     "identity-obj-proxy": "^3.0.0",
     "is-interactive": "^2.0.0",
-    "jest": "^29.7.0",
     "jest-axe": "^9.0.0",
     "jest-chain": "^1.1.6",
```
fengelniederhammer, line 253: Do we still need all those jest dependencies?

web/tests/home.spec.ts
```diff
@@ -0,0 +1,15 @@
...
+
+test('has title', async ({ page }) => {
+  await expect(page).toHaveTitle(/CoVariants/)
+})
+
+test('navbar link', async ({ page }) => {
+  await page.getByRole('link', { name: 'FAQ' }).click()
+
+  await expect(page.getByRole('heading', { name: 'Frequently asked questions' })).toBeVisible()
+})
```
fengelniederhammer, line 15: I usually like to group those into describe block, because it yields nicer descriptions when running the tests and the code is grouped logically.
```suggestion
test.describe('The Home page', () => {
test.beforeEach(async ({ page }) => {
  await page.goto('/')
})

test('has title', async ({ page }) => {
  await expect(page).toHaveTitle(/CoVariants/)
})

test('navbar link', async ({ page }) => {
  await page.getByRole('link', { name: 'FAQ' }).click()

  await expect(page.getByRole('heading', { name: 'Frequently asked questions' })).toBeVisible()
})
});
```

fengelniederhammer (approved):
> vitest for unit tests as the configuration setup seems orders of magnitude easier than for jest.

nice :)

web/playwright.config.ts
fengelniederhammer, line 1: IMO we could also delete all those comments in that file. But no strong opinion, it's up to you.

web/playwright.config.ts
```diff
@@ -0,0 +1,79 @@
...
+  workers: process.env.CI ? 1 : undefined,
+  /* Reporter to use. See https://playwright.dev/docs/test-reporters */
+  reporter: [['list'], ['html', { outputFolder: '.playwright-report' }]],
+  /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
+  use: {
+    /* Base URL to use in actions like `await page.goto('/')`. */
+    baseURL: 'http://127.0.0.1:3000',
+
+    /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
+    trace: 'on-first-retry',
```
fengelniederhammer, line 32: This is usually useful:
```suggestion
        trace: 'retain-on-failure',
        screenshot: 'only-on-failure',
```

web/package.json
AdvancedCodingMonkey: No, I removed them and that even resolved one more console warning, nice!

PR #448, opened 2024-12-10
404 prune package.json

AdvancedCodingMonkey:
This PR further prunes the package.json (scripts and dependencies) and removes babel-node from the execution of the build commands. I think the point there was to support the old node version the project was using but since we moved to the current node version, this is not needed anymore.


web/package.json
```diff
@@ -5,48 +5,35 @@
...
-    "dev:clean": "rimraf '.build/development/{*,.*}' 'node_modules/.cache'",
-    "prod:build": "yarn run next:build && yarn run next:export",
-    "prod:build:ci": "yarn run prod:build",
+    "dev:nowatch": "cross-env NODE_ENV=development BABEL_ENV=development next dev --port 3000",
+    "dev:clean": "rimraf '.build/development/**/*' 'node_modules/.cache'",
+    "prod:build": "yarn next:build && yarn next:export",
     "prod:build:vercel": "cp .env.vercel .env && yarn install && yarn prod:build && cp .next/routes-manifest.json .build/production/web/",
-    "prod:_build": "yarn run next:build && yarn run next:export",
-    "next:build": "cross-env NODE_ENV=production BABEL_ENV=production NODE_OPTIONS=--max_old_space_size=5000  babel-node --config-file \"./babel-node.config.js\" --extensions \".js,.ts\" -- node_modules/.bin/next build",
+    "prod:build:profile": "yarn next:build:profile && yarn next:export",
```
fengelniederhammer, line 14: Should we also inline this one? Or is it useful somehow?

web/package.json
```diff
@@ -5,48 +5,35 @@
...
     "dev": "nodemon --config config/nodemon/dev.json",
     "dev:start": "yarn install && yarn dev:nowatch || cd .",
-    "dev:nowatch": "cross-env NODE_ENV=development BABEL_ENV=development NODE_OPTIONS=--max_old_space_size=5000  babel-node --config-file \"./babel-node.config.js\" --extensions \".js,.ts\" -- node_modules/.bin/next dev --hostname 0.0.0.0 --port 3000",
-    "dev:start_win": "yarn install && yarn dev:nowatch_win || cd .",
-    "dev:nowatch_win": "cross-env NODE_ENV=development BABEL_ENV=development NODE_OPTIONS=--max_old_space_size=5000  babel-node --config-file \"./babel-node.config.js\" --extensions \".js,.ts\" -- ./node_modules/next/dist/bin/next dev --hostname 0.0.0.0 --port 3000",
-    "dev:clean": "rimraf '.build/development/{*,.*}' 'node_modules/.cache'",
-    "prod:build": "yarn run next:build && yarn run next:export",
-    "prod:build:ci": "yarn run prod:build",
+    "dev:nowatch": "cross-env NODE_ENV=development BABEL_ENV=development next dev --port 3000",
+    "dev:clean": "rimraf '.build/development/**/*' 'node_modules/.cache'",
```
fengelniederhammer, line 11: You deleted `prod:clean`. Is this anymore useful in this state?
* Should we delete it, too?
* Should we adapt it to delete `build`, `playwright-report`, ... (i.e. everything that is autogenerated somehow)?

web/package.json
```diff
@@ -5,48 +5,35 @@
...
-    "clear": "node ./lib/clear.js",
     "dev": "nodemon --config config/nodemon/dev.json",
     "dev:start": "yarn install && yarn dev:nowatch || cd .",
-    "dev:nowatch": "cross-env NODE_ENV=development BABEL_ENV=development NODE_OPTIONS=--max_old_space_size=5000  babel-node --config-file \"./babel-node.config.js\" --extensions \".js,.ts\" -- node_modules/.bin/next dev --hostname 0.0.0.0 --port 3000",
-    "dev:start_win": "yarn install && yarn dev:nowatch_win || cd .",
-    "dev:nowatch_win": "cross-env NODE_ENV=development BABEL_ENV=development NODE_OPTIONS=--max_old_space_size=5000  babel-node --config-file \"./babel-node.config.js\" --extensions \".js,.ts\" -- ./node_modules/next/dist/bin/next dev --hostname 0.0.0.0 --port 3000",
-    "dev:clean": "rimraf '.build/development/{*,.*}' 'node_modules/.cache'",
-    "prod:build": "yarn run next:build && yarn run next:export",
-    "prod:build:ci": "yarn run prod:build",
+    "dev:nowatch": "cross-env NODE_ENV=development BABEL_ENV=development next dev --port 3000",
```
fengelniederhammer, line 10: Do we really need `dev`, `dev:start` and `dev:nowatch`?

web/tools/server/server.ts
fengelniederhammer, line 1: Do we actually really need this? Maintaining a server that serves files seems weird to me. There are server that can do that out of the box.

web/package.json
fengelniederhammer, line 1: Do we need:
`lint:watch`,
`lint:fix:watch`,
`eslint:watch`,
`eslint:fix:watch`,
`tsc:project:watch`,
`test:unit:nowatch`
`i18n:extract`
`i18n:addkeys`
`i18n:fix`
`i18n:translate`
?

web/config/nodemon/dev.json
fengelniederhammer, line 1: What does this actually do?

According to a quick google search this is some sort of hot reloading config for `nodemon` - Can't we simply get rid of it? Isn't next capable of something like this (I suppose every build tool in the JS world is?)?

web/lib/EnvVarError.js
fengelniederhammer, line 1: This whole `web/lib` seems suspicious to me - maybe we can get rid of that somewhen later? Or at least clean it up a bit (the weird mixture of `.js`, `.ts` and `.d.ts`)

AdvancedCodingMonkey: Right, I refactored to only using .ts files. I also find it a bit suspicious / annoying; I think the opportunities to remove parts will present themselves when we touch other things later.

web/package.json
AdvancedCodingMonkey: For dev, I like having the nodemon watch because it triggers on more files than next's hot reload (e.g. when adding packages). Inlining `dev:nowatch` is nice now that the command is shorter.

web/package.json
AdvancedCodingMonkey: The translation commands we need. The watch commands we don't need, all of those things the IDE should do.

web/package.json
AdvancedCodingMonkey: I just moved the `prod:clean` to a different line but you are right, as the `watch` scripts are gone it is not needed anymore. Overall, caches and autogenerated things are maintained by the tools that create them, so we really don't need a script to clean them every time. I will remove both `clean` scripts.

web/config/nodemon/dev.json
AdvancedCodingMonkey: It does hot reloading when files change. For `dev` I find it useful because it watches files that `next` does not watch by default. For the other contexts, IDE integration takes replaces using something like `nodemon`, so I removed those scripts.

Issue #27, opened 2021-01-19
Remember filter state on page navigation


ivan-aksamentov:
Currently filters on sidebar of plot pages are reset on every navigation, This is dues to state being stored locally.

Moving the state upstream, or better, to redux store would make it persist.

AdvancedCodingMonkey:
Implemented as drive-by in PR #478

PR #478, opened 2025-01-23
103 reduce rerendering

AdvancedCodingMonkey:
Refactor components and state to reduce re-rendering (e.g. whole page re-renders when country / cluster filters are used). State refactorings are based around moving the initial fetch to a recoil atom and then using selectors to get derived state. As a drive-by, this means filter state is now preserved _per page_ when navigating around (closes issue #27 ).


web/src/components/DistributionSidebar/CountryFilters.tsx
```diff
@@ -58,8 +59,9 @@ export function IconComponent({ country, Icon }: IconOrLineComponentProps) {
 export function LineComponent({ country }: IconOrLineComponentProps) {
   const { t } = useTranslationSafe()
   const theme = useTheme()
-  const stroke = useCountryColor()(country)
-  const strokeDasharray = useCountryStrokeDashArray()(country)
+  const getCountryStyles = useRecoilValue(getCountryStylesSelector)
+  const stroke = getCountryStyles(country).color
+  const strokeDasharray = getCountryStyles(country).strokeDashArray
```
fengelniederhammer, line 64:
```suggestion
  const { color: stroke, strokeDashArray: strokeDasharray } = getCountryStyles(country)
```

web/src/io/getPerClusterData.ts
```diff
@@ -1,70 +1,34 @@
...
+  country_names: z.string().array(),
+  distributions: clusterDistributionSchema.array(),
+})
+
+export type PerClusterDataRaw = z.infer<typeof perClusterDataRawSchema>
+export type ClusterDistribution = z.infer<typeof clusterDistributionSchema>
+export type ClusterDistributionDatum = z.infer<typeof clusterDistributionDatumSchema>
+
+export async function fetchPerClusterDataRaw() {
+  const data = await FETCHER.fetch<PerClusterDataRaw>('/data/perClusterData.json')
```
fengelniederhammer, line 30: We don't really need the explicit type, since we use the zod schema to parse `data` anyway?
```suggestion
  const data = await FETCHER.fetch('/data/perClusterData.json')
```

web/src/io/getPerCountryCasesData.ts
```diff
@@ -41,38 +40,20 @@ const perCountryCaseDataRawSchema = z.object({
...
+export type PerCountryCasesDataRaw = z.infer<typeof perCountryCaseDataRawSchema>
 type PerCountryCasesDatum = z.infer<typeof perCountryCasesDatumSchema>
 export type PerCountryCasesDistribution = z.infer<typeof perCountryCasesDistributionSchema>
 export type PerCountryCasesDistributionDatum = z.infer<typeof perCountryCasesDistributionDatumSchema>
 
-export function fetchPerCountryCasesDataRaw(): Promise<PerCountryCasesDataRaw> {
-  return FETCHER.fetch<PerCountryCasesDataRaw>('/data/perCountryDataCaseCounts.json')
+export async function fetchPerCountryCasesDataRaw(): Promise<PerCountryCasesDataRaw> {
+  const data = await FETCHER.fetch<PerCountryCasesDataRaw>('/data/perCountryDataCaseCounts.json')
+  return perCountryCaseDataRawSchema.parse(data)
```
fengelniederhammer, line 50: Looks like a pattern... :) We could add a method to the `FETCHER` that takes a url and a schema.

web/src/state/utils/atomDefault.ts
```diff
@@ -0,0 +1,41 @@
...
+  key: string
+  effects?: ReadonlyArray<AtomEffect<T>>
+  effects_UNSTABLE?: ReadonlyArray<AtomEffect<T>>
+  dangerouslyAllowMutability?: boolean
+  default: (opts: { get: GetRecoilValue; getCallback: GetCallback }) => T
+}
+
+/* Atom with `default` accepting also a function */
+export function atomDefault<T>(options: AtomDefaultOptions<T>): RecoilState<T> {
+  if (typeof options.default === 'function') {
```
fengelniederhammer, line 14: How is this ever not a function? Doesn't the type already guarantee that?

web/src/state/utils/atomDefault.ts
```diff
@@ -0,0 +1,41 @@
+import type { AtomEffect, GetCallback, GetRecoilValue, RecoilState, SerializableParam } from 'recoil'
+import { atom, atomFamily, selector, selectorFamily } from 'recoil'
+
+export interface AtomDefaultOptions<T> {
+  key: string
+  effects?: ReadonlyArray<AtomEffect<T>>
+  effects_UNSTABLE?: ReadonlyArray<AtomEffect<T>>
+  dangerouslyAllowMutability?: boolean
+  default: (opts: { get: GetRecoilValue; getCallback: GetCallback }) => T
+}
```
fengelniederhammer, line 10: This is essentially the type of the argument of `atom` except that we modify `default`, isn't it? We could also express that as a type more directly:
```suggestion
export type AtomDefaultOptions<T> = Omit<AtomOptions<T>, 'default'> & {
  default: (opts: { get: GetRecoilValue; getCallback: GetCallback }) => T
}
```

fengelniederhammer (approved):
LGTM :+1:

web/src/io/getPerCountryCasesData.ts
AdvancedCodingMonkey: That is exactly what I had in mind, just hadn't gotten around to doing it yet :)

web/src/state/utils/atomDefault.ts
AdvancedCodingMonkey: Nice suggestion!

web/src/state/utils/atomDefault.ts
AdvancedCodingMonkey: You are right but I think there is a mistake here. A function should be allowed in addition to other possibilities, it should not be exclusive. I will have to fix that.

PR #484, opened 2025-01-30
feat-287: remove 0-values from tooltips

AdvancedCodingMonkey:
Remove countries / variants from tooltips that have value 0


fengelniederhammer (approved):
Nice :)

PR #495, opened 2025-02-06
Create a staging environment for pull requests

AdvancedCodingMonkey:


AdvancedCodingMonkey:
The idea here is to have a staging deployment that gets triggered from any PR. To reduce deployment time, I wanted to reduce the amount of copying files to the s3 bucket. There seem to be feature requests for [checksum-based syncing with aws,](https://github.com/aws/aws-cli/issues/9074) which would fit the need exactly, but the feature is not implemented yet. I took inspiration from [this custom solution](https://github.com/espressif/developer-portal/pull/381/files) to write our own custom solution.

.github/workflows/web_staging.yml
```diff
@@ -0,0 +1,207 @@
...
+          if aws s3 ls s3://${AWS_S3_BUCKET}/checksums-ack-s3.txt; then
+            echo "s3 checksums found, downloading s3 checksums"
+            aws s3api get-object --bucket ${AWS_S3_BUCKET} --key checksums-ack-s3.txt checksums-ack-s3.txt
+          else
+            echo "no s3 checksums found, using ci checksums"
+            touch checksums-ack-s3.txt
+          fi
+          echo "comparing checksums"
+          comm -23 checksums-ack-ci.txt checksums-ack-s3.txt | awk '{print $1}' > outdated_acknowledgements.txt
+          echo "found outdated files"
```
fengelniederhammer, line 166: This is maybe a bit confusing, since it will also say "found outdated files" if it didn't find any.
```suggestion
          echo "Listing outdated files:"
```

.github/workflows/web_staging.yml
```diff
@@ -0,0 +1,207 @@
...
+          cat outdated_acknowledgements.txt
+          comm -13 checksums-ack-ci.txt checksums-ack-s3.txt | awk '{print $1}' | comm -23 - outdated_acknowledgements.txt > obsolete_acknowledgements.txt
+          echo "found obsolete files"
```
fengelniederhammer, line 169: (same here)

.github/workflows/web_staging.yml
```diff
@@ -0,0 +1,207 @@
...
+      - name: "[init] Checkout code"
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+          submodules: true
+
+      - name: "[build] Setup environment (staging)"
+        run: |
+          echo "ENV_NAME=staging" >> $GITHUB_ENV
+          echo "FULL_DOMAIN=https://covariants.org" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
AdvancedCodingMonkey: It would seem so (Ivan set it up this way). FULL_DOMAIN is always only the covariants part, the subdomains are not needed anywhere. I think it might have been different in the past, when circleCI was used and the whole thing was hosted via vercel. Now, the link to staging, master, release comes solely from which bucket is written to, which is determined by the job environment.

AdvancedCodingMonkey:
I have reduced code duplication by creating composite actions that can be used in both the staging and the prod workflow

.github/workflows/build_app/action.yml
```diff
@@ -0,0 +1,57 @@
...
+        cd web
+        yarn install --frozen-lockfile
+      shell: bash
+
+    - name: "[build] Build web application"
+      run: |
+        echo "[build] Build web application"
+        cd web
+        yarn prod:build
+      shell: bash
```
fengelniederhammer, line 57: Tailing newlines wouldn't hurt (that's why GitHub shows that red round icon)

fengelniederhammer (approved):
LGTM

PR #519, opened 2025-02-20
Update alt display names to always be pango

emmahodcroft:
Updates so that `alt_display_name` is always the Pango lineage, to make this an easy 'alternative' for dispalying instead of the Nextclade name!


AdvancedCodingMonkey:
Thanks for the update @emmahodcroft , these variants seem to have no `alt_display_name` at all but they have pango lineages. Could you update those as well?
- 21I (Delta)
- 21J (Delta)
- 21G (Lambda)
- 21H (Mu)
- 20B/S:732A
- 20A/S:126A
- 20A/S:439K
- 20A.EU2
- 20A/S:98F
- 20A/S:439K
- 20C/S:80Y
- 20B/S:626S
- 20B/S:1122L
Also: 20A.EU2 does not seem to have a nextclade name (shouldn't it be 20E (EU2?). If so, could you update the `display_name` field for this variant as well?

emmahodcroft:
> Also: 20A.EU2 does not seem to have a nextclade name (shouldn't it be 20E (EU2?). If so, could you update the display_name field for this variant as well?

Unfortunately 20A.EU2 doesn't have an official Nextclade designation. Is the issue just that the name isn't fitting a format needed for display, or is there a deeper issue at play?

PR #546, opened 2025-03-11
feat-545: switch from using pango_lineages to alt_display_name

AdvancedCodingMonkey:
Change from the intermittent use of `pango_lineages` (due to missing data) to using `alt_display_name` for toggling between nomenclatures.


JonasKellerer (approved):
LGTM

Issue #472, opened 2025-01-15
Per-country page: using country / continent filter spills into "United States" region


AdvancedCodingMonkey:
Reproduce 

1. navigate to /per-country
2. deselect a continent or country
3. navigate to "United States" region
4. only "Georgia" is selected

Desired behavior: all states are selected

Only occurs when one uses the filter elements for the first time

AdvancedCodingMonkey:
Actually, the spill goes for both regions. In "United States", only "United States" and "Gergia" is selected, in "Switzerland", nothing is selected.

PR #548, opened 2025-03-11
472 fix country filters

AdvancedCodingMonkey:
Visiting /per-country, using the country filters and then switching to a different region would leave the "country" filter for that region almost empty (see #472). 

The reason was that the default state of a country filter is taken from the url query if one is present. Using the filter in one region adds a query to the url. Switching regions updates the query but only *after* the filter state has been updated because the filter state depends directly on the selected region state. Consequently, the filter state for e.g. "Switzerland" would be taken from the previous query of e.g. "World", leaving no regions selected.

This PR fixes the bug by only setting the filter state from the url if it includes the correct region.

Bug:

[Screencast from 11.03.2025 11:47:41.webm](https://github.com/user-attachments/assets/cd50de30-9bf5-4090-8de0-e07a6367f58d)

Fixed:

[Screencast from 11.03.2025 11:48:51.webm](https://github.com/user-attachments/assets/d04fe6f3-b2ec-423c-ad78-1f43990c2306)

### DIFF
--- a/web/src/state/Places.ts
+++ b/web/src/state/Places.ts
@@ -2,9 +2,11 @@ import { z } from 'zod'
 import { FETCHER } from 'src/hooks/useAxiosQuery'
 import { atomAsync } from 'src/state/utils/atomAsync'
 
-export const WHOLE_WORLD_REGION = 'World'
-export const UNITED_STATES_REGION = 'United States'
-export const DEFAULT_REGION = WHOLE_WORLD_REGION
+export const REGIONS = {
+  WORLD: 'World' as const,
+  UNITED_STATES: 'United States' as const,
+}
+export const DEFAULT_REGION = REGIONS.WORLD
 
 export interface Country {
   country: string
@@ -35,7 +37,7 @@ export const regionCountryAtom = atomAsync<RegionCountry>({
 })
 
 export const getAllContinents = (region?: string, regionCountry?: RegionCountry) => {
-  if (region === WHOLE_WORLD_REGION && regionCountry !== undefined) {
+  if (region === REGIONS.WORLD && regionCountry !== undefined) {
     return Object.keys(regionCountry).map((continent) => ({ continent, enabled: true }))
   }
   return [{ continent: region ?? '', enabled: true }]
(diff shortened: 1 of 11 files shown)


web/src/state/Places.ts
```diff
@@ -3,6 +3,7 @@ import { FETCHER } from 'src/hooks/useAxiosQuery'
 import { atomAsync } from 'src/state/utils/atomAsync'
 
 export const WHOLE_WORLD_REGION = 'World'
+export const UNITED_STATES_REGION = 'United States'
```
JonasKellerer, line 6: Similar to the other PR: It would be nice to have the regions also typed. This would also get rid of the "_REGION" identifier.

web/src/state/PlacesForPerCountryData.ts
```diff
@@ -51,12 +51,15 @@ export const perCountryCountriesAtom = atomFamilyDefault<Country[], string>({
...
     const distributions = get(perCountryDataDistributionsSelector(region))
     const countries = distributions.map(({ country }) => ({ country, enabled: true }))
-
-    const enabledCountries = convertToArrayMaybe(getLodash(query, 'country'))
-    if (enabledCountries) {
+    const regionFromUrl = getLodash(query, 'region')
+    const enabledCountriesFromUrl = convertToArrayMaybe(getLodash(query, 'country'))
+    // we have to check that the regions are the same here, otherwise the query can spill from one region to another
+    // because the atom effect that updates the query (setRegionAtom) is run *after* all dependent atoms are updated
+    // see issue https://github.com/hodcroftlab/covariants/issues/472
```
JonasKellerer, line 58: Can we add a test, so that we don't need this comment?

JonasKellerer (approved):
Nice find. The fix works for me. 
I would advocate for a test, so we don't need the comment, which can be deprecated. However, I am not sure if the current setup allows that?

web/src/state/PlacesForPerCountryData.ts
AdvancedCodingMonkey: Puh, that took me longer than I would have liked. Yes, we now have a test :)

PR #549, opened 2025-03-11
465 fix missing countries

AdvancedCodingMonkey:
Countries are missing from the region-country mapping and thus not properly selected via the country filters.

### DIFF
--- a/web/src/components/DistributionSidebar/CountryFilters.tsx
+++ b/web/src/components/DistributionSidebar/CountryFilters.tsx
@@ -93,8 +93,8 @@ export function CountryFilterCheckbox({
 
   return (
     <FormGroup check>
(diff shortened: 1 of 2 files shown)


web/src/components/DistributionSidebar/CountryFilters.tsx
```diff
@@ -96,7 +93,7 @@ export function CountryFilterCheckbox({
 
   return (
     <FormGroup check>
-      <Label htmlFor={CSS.escape(country)} check>
+      <Label className={'d-flex align-items-baseline gap-2'} htmlFor={CSS.escape(country)} check>
```
JonasKellerer, line 96: I think we don't need the htmlFor here, since the input element is contained in the label object. But I am not sure on that. This is at least the case for standart `<label>` and `<input>` elements.

JonasKellerer (approved):
LGTM

PR #550, opened 2025-03-11
fix-517: make distribution sidebar states global and introduce jotai

AdvancedCodingMonkey:
Sidebar menu states reset when navigating through the app. This PR puts them into the global state and thus makes them persistent until page refresh. Since we want to move away from recoil, I took the chance to introduce it so we already have working alternative we can expand on.


web/src/components/Cases/CasesPage.tsx
```diff
@@ -37,8 +41,10 @@ function CasesPlotSection() {
           setCountries={setCountries}
           regionsTitle={t('Countries')}
           enabledFilters={enabledFilters}
-          clustersCollapsedByDefault={false}
-          countriesCollapsedByDefault={false}
+          clustersCollapsed={clustersCollapsed}
+          setClustersCollapsed={setClustersCollapsed}
+          countriesCollapsed={countriesCollapsed}
```
JonasKellerer, line 46: I think we dont have to pass this variables down to the components. We could just use the atom directly where it is used, eg. directly in the `CountryFilter`. (This is my understanding, why we have the atoms in the first place).

We might also think about to only optionally pass the atom name to the `CardCollapsible`, which stores the state. Then the card would be responsible of knowing, if it is open or not. This would make it easier to test.

It would be nice to start with tests directly using the new framework. Then we can add a test each time we use the new system.

web/package.json
```diff
@@ -53,6 +53,7 @@
     "intersection-observer": "^0.12.2",
     "is-absolute-url": "^4.0.1",
     "iso-3166-1-alpha-2": "^1.0.2",
+    "jotai": "2.12.2",
```
JonasKellerer, line 56: I have no clue about state libraries, so I have to ask: What makes this a good choice?

JonasKellerer (changes requested):
Works like a charm :)

I would put the responsiblity of the state closer to the the component that uses it.

web/package.json
AdvancedCodingMonkey: It was between this and zedux. Zedux sounded a bit less straightforward in the docs. Jotai was recommended in some github discussions about the recoil no longer being maintained. Since I also have no experience with either, I thought I'd try out jotai and see how far it takes us.

web/src/components/Cases/CasesPage.tsx
AdvancedCodingMonkey: Nice suggestion. After our discussion, I implemented handing down the atoms and then using them as close to the component as possible. Since `CardCollapsible` is used both with global and local state, I decided for it to have state value and setter as props. This allows flexible use without adding use-case-specific decision logic to the component itself.

web/src/components/Common/CardCollapsible.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+}
+
+export const UseLocalState: Story = {
+  render: (args) => {
+    return <LocalStateWrapper {...args} />
+  },
+  args: {
+    title: 'Collapse This!',
+    children: 'Expaaaanded',
+  },
```
JonasKellerer, line 42: I think we dont need this, since we already provide the default args in the meta object.

web/src/components/Common/CardCollapsible.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+  play: async ({ canvasElement }) => {
+    const canvas = within(canvasElement)
+
+    const expandable = canvas.getByText('Collapse This!')
+    await userEvent.click(expandable)
+    await userEvent.click(expandable)
```
JonasKellerer, line 18: We could check in between, that the "Expanded" is visible. and after clicking again it is now longer visible.

JonasKellerer (commented):
Nice. I like it.
Sadly, I could not run the storybook. I get this error:
```
SB_BUILDER-WEBPACK5_0003 (WebpackCompilationError): There were problems when compiling your code with Webpack.
Run Storybook with --debug-webpack for more information.
```
Did you encounter the same issue?

web/src/components/Common/CardCollapsible.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+import React, { PropsWithChildren, useState } from 'react'
+import { atom, useAtom } from 'jotai'
+import { userEvent, within } from '@storybook/test'
+import { CardCollapsible, CollapsibleCardProps } from 'src/components/Common/CardCollapsible'
+
+const meta: Meta<typeof CardCollapsible> = {
+  component: CardCollapsible,
+  args: {
+    title: 'Collapse This!',
+    children: 'Expaaaanded',
```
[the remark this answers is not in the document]
AdvancedCodingMonkey: :crying_cat_face:

AdvancedCodingMonkey:
> Nice. I like it. Sadly, I could not run the storybook. I get this error:
> 
> ```
> SB_BUILDER-WEBPACK5_0003 (WebpackCompilationError): There were problems when compiling your code with Webpack.
> Run Storybook with --debug-webpack for more information.
> ```
> 
> Did you encounter the same issue?

No, for me it works fine, even after deleting `node_modules` and reinstalling everything. But I noticed that storybook complains about not finding webpack
image
Perhaps for you storybook finds it and that's where the error comes in?

PR #556, opened 2025-03-18
fix-518: fix and document translations

AdvancedCodingMonkey:
Fix errors thrown by the translation scripts and document how to keep the translations up to date.

### DIFF
--- a/docs/Internationalization.md
+++ b/docs/Internationalization.md
@@ -1,0 +1,34 @@
+# Internationalization
+
+Our goal is to make CoVariants accessible to as many people as possible. We provide (auto-generated) translation for interactive elements of the website. 
+Providing translations for all content is unfortunately currently beyond our resources. If you would like to help and provide a translation for a language
+you are familiar with, feel free to make a pull request.
+
+## Generating translations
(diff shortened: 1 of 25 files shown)


docs/Internationalization.md
```diff
@@ -0,0 +1,34 @@
...
+
+```bash
+yarn i18n:fix
+```
+Check keys and apply some fixes.
+
+```bash
+yarn i18n:translate
+```
+Translate all keys from English to the languages found in `i18n/resources`. Needs valid aws credentials.
```
JonasKellerer, line 30: Should we add a github action, that runs all these steps and commits the output to a specific branch? This way we can store the aws credentials in the github secrets and every maintainer can run it. We have a simliar action in genspectrum, see https://github.com/GenSpectrum/dashboard-components/blob/main/.github/workflows/update-snapshots.yml

JonasKellerer (approved):
LGTM. The translations work for me. I could no test the workflow itself, since I have not the aws credentials.

docs/Internationalization.md
AdvancedCodingMonkey: Thanks, that would be nice. Ivan pointed out that the aws translation service is rather expensive though, which might prohibit something that anybody can run at any time. I will discuss this further and possible add it in a separate PR.

PR #565, opened 2025-03-19
feat-521: enhance design of defining mutations tables

### LINKED ISSUE #521: Defining mutations page: Design mutation table
Investigate a new design of the mutations table. The table could include pagination, searching. Also dont show the comments of the mutations so promiently.

JonasKellerer:
Resolves #521

Improves the design of the defining mutations table.

Uses tanstack table for table functionality and bootstrap table for styling.
The data can be filtered and sorted. The table has pagination.

The table of coding and silent mutations are next to each other. On small screens they move to a column design. 

The annotations are now behind an info icon. When the user hovers over the icon, the annotation is shown.

grafik


web/src/components/Common/table/TableWithSearchPaginationFilter.tsx
```diff
@@ -0,0 +1,128 @@
...
+            <tr key={row.id}>
+              {row.getVisibleCells().map((cell) => (
+                <td key={cell.id}>{flexRender(cell.column.columnDef.cell, cell.getContext())}</td>
+              ))}
+            </tr>
+          ))}
+          {table.getRowModel().rows.length === 0 && (
+            <tr>
+              <td colSpan={2} className={'p-2'}>
+                Nothing to show
```
AdvancedCodingMonkey, line 50: I find this text a bit too large

web/src/components/Common/table/TableWithSearchPaginationFilter.tsx
```diff
@@ -0,0 +1,128 @@
...
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function Filter({ column }: { column: Column<any> }) {
+  const columnFilterValue = column.getFilterValue()
+
+  const onChange = useCallback((value: string | number) => column.setFilterValue(value), [column])
+
+  return (
+    <DebouncedInput
```
AdvancedCodingMonkey, line 91: I don't quite get what "debouncing" is

web/src/components/Common/table/TableWithSearchPaginationFilter.tsx
```diff
@@ -0,0 +1,128 @@
+import { Column, flexRender, Header, SortDirection } from '@tanstack/react-table'
+import React, { useCallback, useEffect, useState } from 'react'
+import type { Table } from '@tanstack/table-core'
+import { Pagination } from 'src/components/Common/table/Pagination'
+
+export function TableWithSearchPaginationFilter({ table, pageSizes }: { table: Table<any>; pageSizes: number[] }) {
```
AdvancedCodingMonkey, line 6: Nice job on this table! Can we resolve the types a bit more to remove the `any`s? I'd be happier if we just allow the data types for the two tables rather than using any.

web/src/components/Common/table/Pagination.tsx
```diff
@@ -0,0 +1,153 @@
+import React, { ChangeEvent, useCallback } from 'react'
+import type { Table } from '@tanstack/table-core'
+
+interface PaginationProps {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  table: Table<any>
```
AdvancedCodingMonkey, line 6: Same here as below, can we make this type more explicit? Even if it is just the union of our two tableData types?

web/src/components/Common/table/Pagination.tsx
```diff
@@ -0,0 +1,153 @@
...
+  const onChange = useCallback(
+    (event: ChangeEvent<HTMLInputElement>) => {
+      const page = event.currentTarget.value ? Number(event.currentTarget.value) - 1 : 0
+      table.setPageIndex(page)
+    },
+    [table],
+  )
+
+  if (table.getRowModel().rows.length === 0) {
+    return null
```
AdvancedCodingMonkey, line 88: Hmm, if I type something in the search and nothing comes up, the `GoToPageSelector` disappears. That is fine. But if I hit a number in the selector that exceeds the pages we have (e.g. 10), the selector also disappears. Can we make it such that it will go to 8 instead or do nothing?

AdvancedCodingMonkey: Oh, and internally, the counter does go to page 10, so if I hit the "go back a page"  button twice, I return to page 8. When there are no search results however, the page navigation buttons are disabled.

web/src/components/Common/table/Pagination.tsx
```diff
@@ -0,0 +1,153 @@
...
+        ))}
+      </select>
+    </label>
+  )
+}
+
+function GotoPageSelector({ table }: PaginationProps) {
+  const onChange = useCallback(
+    (event: ChangeEvent<HTMLInputElement>) => {
+      const page = event.currentTarget.value ? Number(event.currentTarget.value) - 1 : 0
```
AdvancedCodingMonkey, line 81: This immediately tracks the changes made in the selector; meaning if I hit backspace in there, I immediately jump a to a different page. I find it a bit jumpy. Perhaps the selection should only be updated when hitting enter? Feel free to ignore my opinion here though if you disagree or it would be a hassle to fix.

web/src/components/Common/table/Pagination.tsx
```diff
@@ -0,0 +1,153 @@
...
+  )
+}
+
+function PageSizeSelector({
+  table,
+  pageSizes,
+}: PaginationProps & {
+  pageSizes: PageSizes
+}) {
+  const onChange = useCallback(
```
AdvancedCodingMonkey, line 48: Hmm, when setting the page size, the `GoToPageSelector` is not updated and can display a wrong page.

web/src/components/Common/table/TableWithSearchPaginationFilter.tsx
```diff
@@ -0,0 +1,128 @@
...
+function Filter({ column }: { column: Column<any> }) {
+  const columnFilterValue = column.getFilterValue()
+
+  const onChange = useCallback((value: string | number) => column.setFilterValue(value), [column])
+
+  return (
+    <DebouncedInput
+      className="form-control form-control-sm"
+      onChange={onChange}
+      placeholder={`Search...`}
```
AdvancedCodingMonkey, line 94: A button in the search field to remove the search term would be nice.

web/src/components/DefiningMutations/DefiningMutationByCluster/DefiningMutationsTables.tsx
```diff
@@ -0,0 +1,272 @@
...
+    debugTable: true,
+  })
+
+  return <TableWithSearchPaginationFilter table={table} pageSizes={pageSizes} />
+}
+
+function Annotation({ annotationText }: { annotationText: string }) {
+  return (
+    <MutationBadgeBox className={'text-info'} data-bs-toggle="tooltip" title={annotationText}>
+      <span aria-label="Annotation" className="bi bi-info-circle"></span>
```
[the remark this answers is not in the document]
AdvancedCodingMonkey: Additionally, I find the font of the `bi-info-circle` a bit strange. Can one easily change that?

web/src/components/DefiningMutations/DefiningMutationByCluster/DefiningMutationsTables.tsx
```diff
@@ -0,0 +1,272 @@
...
+    getFilteredRowModel: getFilteredRowModel(),
+    getPaginationRowModel: getPaginationRowModel(),
+    initialState: {
+      sorting: [{ id: 'nucleotideMutations', desc: false }],
+    },
+    onPaginationChange: setPagination,
+    state: {
+      pagination,
+    },
+    debugTable: true,
```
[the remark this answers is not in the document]
JonasKellerer: Good catch. No we don't need this.

web/src/components/DefiningMutations/DefiningMutationByCluster/DefiningMutationsTables.tsx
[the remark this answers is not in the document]
JonasKellerer: I don't think so. This is a fixed image/svg. If you have another suggestion for an icon, I am happy to use that. I just typed "info" in the search of bootstrap icons: https://icons.getbootstrap.com/?q=info

web/src/components/DefiningMutations/DefiningMutationByCluster/DefiningMutationsTables.tsx
```diff
@@ -0,0 +1,272 @@
...
+          }
+          return firstNucMutA.pos - firstNucMutB.pos
+        },
+        filterFn: (row, _, filterValue: string) => {
+          const mutationString = formatMutation(getMutationFromNucleotideMutation(row.original.nucleotideMutation))
+
+          return mutationString.includes(filterValue)
+        },
+      }),
+    ]
```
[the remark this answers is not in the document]
JonasKellerer: Not entirely. The one above takes a row which contains a list of nucleotide mutations and displays them, and sorts for the first entry and runs the filter on the list. Here this is only one mutations, since they are no longer grouped to one amino acid mutation.

web/src/components/Common/table/TableWithSearchPaginationFilter.tsx
```diff
@@ -0,0 +1,128 @@
...
+import type { Table } from '@tanstack/table-core'
+import { Pagination } from 'src/components/Common/table/Pagination'
+
+export function TableWithSearchPaginationFilter({ table, pageSizes }: { table: Table<any>; pageSizes: number[] }) {
+  return (
+    <div className="border border-2 rounded shadow-sm">
+      <table className={'table table-striped table-hover'}>
+        <thead>
+          {table.getHeaderGroups().map((headerGroup) => (
+            <React.Fragment key={headerGroup.id}>
```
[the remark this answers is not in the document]
JonasKellerer: Yeah. I thought so to. Tobias had the same issue at some point and took a long time to find this.

web/src/components/Common/table/TableWithSearchPaginationFilter.tsx
JonasKellerer: We had the same issue in genspectrum. Usually one would want to use `unkown` here, since the type of the table is generally known, but not when writing this component. However, due to some unknown problem with the inner typings of tanstack table, we can not use `unknown` here. So we could either use any or like you suggested writing the explicit type of each usage of the table. However, I imagine we can reuse this table also in other places and it would be nice not to have to change the type, when there is a new consumer. 

I changed it below, whereever I can for the HeaderText and Filter.

web/src/components/Common/table/Pagination.tsx
JonasKellerer: I could change it here to `unknown` since we only use it in the `TableWithSearchPaginationFilter` so far. If this changes we might need to change it back to any.

web/src/components/Common/table/Pagination.tsx
JonasKellerer: I used the `DebouncedInput` to fix this issue and added a guard, that when there is nothing entered, nothing happens. It now also shows an error, if there is a wrong input.

web/src/components/Common/table/Pagination.tsx
JonasKellerer: I have now used the `DebouncedInput` we also have for the table.

web/src/components/Common/table/Pagination.tsx
JonasKellerer: I think with my changed above, this should no longer be the case.

web/src/components/Common/table/TableWithSearchPaginationFilter.tsx
JonasKellerer: I tried a little bit around and did not find an easy solution so far to integrate the x nicely in the input. I will leave this one open, since this has no high priority in my mind.

web/src/components/DefiningMutations/DefiningMutationByCluster/DefiningMutationsTables.tsx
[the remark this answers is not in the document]
AdvancedCodingMonkey: Right, I mixed that up.

web/src/components/DefiningMutations/DefiningMutationByCluster/DefiningMutationsTables.tsx
```diff
@@ -0,0 +1,272 @@
...
+        },
+        filterFn: (row, _, filterValue: string) => {
+          const mutationStrings = row.original.nucleotideMutations.map((mutation) =>
+            formatMutation(getMutationFromNucleotideMutation(mutation)),
+          )
+          return mutationStrings.some((mutation) => mutation.includes(filterValue))
+        },
+      }),
+    ]
+  }, [])
```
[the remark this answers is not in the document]
JonasKellerer: I now extracted the columns and put each of the tables in their own file. Hopefully this helps. I am not sure if it helps further to extract the cell, sortingFn and filterFn into a component. I would leave it like this for now.

web/src/components/Common/table/TableWithSearchPaginationFilter.tsx
AdvancedCodingMonkey: We can also close it, as you say it does not have a high priority

web/src/components/DefiningMutations/DefiningMutationByCluster/DefiningMutationsTables.tsx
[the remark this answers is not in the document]
AdvancedCodingMonkey: I like this a lot, thanks!

AdvancedCodingMonkey (approved):
Very nice

PR #569, opened 2025-03-27
523 defining mutations file format

AdvancedCodingMonkey:
This PR prepares a script for merging the the auto-generated and hand-curated defining mutations files. In doing so, it also defines the data format that will be used for the defining mutations.

The data format can be inspected in `tests/data/defining_mutations/expected_output/`


scripts/merge_defining_mutations.py
```diff
@@ -0,0 +1,355 @@
...
+        .filter(pl.col('mutation_type') == 'coding')
+        .group_by('lineage', 'nextstrain_clade', 'relative_to', 'mutation_type', 'aa_change')
+        .agg(pl.col('nuc_change'), pl.col('notes').first())
+    )
+    silent_grouped_by_aa = typed.filter(pl.col('mutation_type') == 'silent').with_columns(
+        pl.col('nuc_change').cast(pl.List(pl.String)))
+    grouped_by_aa = pl.concat([
+        coding_grouped_by_aa,
+        silent_grouped_by_aa
+    ])
```
JonasKellerer, line 284: It would be nice to have this in a function.

scripts/merge_defining_mutations.py
```diff
@@ -0,0 +1,355 @@
...
+    deletions_to_ranges = split_deletions.applymap(
+        lambda x: [list(range(*del_range)) if len(del_range) > 1 else del_range for del_range in x])
+    flatten_deletions = deletions_to_ranges.applymap(lambda x: list(pd.core.common.flatten(x)))
+    expand_deletions = empty_strings_removed.copy()
+    expand_deletions[deletion_columns] = flatten_deletions
+
+    reformat_deletions = pl.from_pandas(expand_deletions).with_columns(
+        pl.col(col_name).list.eval(
+            pl.concat_str(
+                pl.lit('X'),
```
JonasKellerer, line 130: I am not sure if we need the `X` symbol here. It least for cov-spectrum, a mutation with `123-` is still a valid mutation. We would need to take this into account for the nucleotide mutations objects above. 
However, how I see it, this will end up in the output. `X` is no ambigous symbol for nucleotides but for amino acids.

scripts/merge_defining_mutations.py
```diff
@@ -0,0 +1,355 @@
...
+                if not silent_mutation['notes']:
+                    silent_mutation.pop('notes')
+            if not mutation_types['coding']:
+                mutation_types['coding'] = []
+            for coding_mutation in mutation_types['coding']:
+                coding_mutation.update({'aa_change': aa_string_to_object(coding_mutation['aa_change'])})
+                coding_mutation.update(
+                    {'nuc_change': [nuc_string_to_object(mutation) for mutation in coding_mutation['nuc_change']]})
+                if not coding_mutation['notes']:
+                    coding_mutation.pop('notes')
```
JonasKellerer, line 335: Could we do this in two steps, first we transform everything to a dict, then remove all empty fields? It is hard to find out if this function does what we want it to from looking at it.

JonasKellerer: We could then also put both steps into its own function :)

scripts/merge_defining_mutations.py
```diff
@@ -0,0 +1,355 @@
...
+         output_dir='../tests/data/defining_mutations/output'):
+    lineages, emma, corn = import_file_dfs(emma_dir, corn_dir)
+
+    merged = merge_file_dfs(emma, corn)
+
+    output = reformat_df_to_dicts(merged, lineages)
+
+    for lineage in output:
+        with open(os.path.join(output_dir, f'{lineage["lineage"]}.json'), 'w') as f:
+            json.dump(lineage, f)
```
JonasKellerer, line 350: I think this could also be put into a function.

scripts/merge_defining_mutations.py
```diff
@@ -0,0 +1,355 @@
...
+
+def nuc_to_gene(nuc_location: int):
+    genes = []
+    for gene, bounds in GENE_BOUNDS.items():
+        if bounds[0] <= nuc_location <= bounds[1]:
+            genes.append(gene)
+    return genes
+
+
+def match_nuc_to_aas(nuc, aas):
```
JonasKellerer, line 41:
```suggestion
def match_nuc_to_aas(nuc: str, aas: list[str]) -> str | None:
```

scripts/merge_defining_mutations.py
```diff
@@ -0,0 +1,355 @@
...
+        ),
+        aa_change_pango_parent=pl.concat_list(
+            'aa_sub_pango_parent',
+            'aa_del_pango_parent'
+        )
+    )
+
+    # TODO: make sure the nuc to aa matching is scientifically correct
+    match_aa_changes = pl.struct('nuc_change', 'aa_change').map_elements(
+        lambda x: match_nuc_to_aas(x['nuc_change'], x['aa_change']), pl.String
```
JonasKellerer, line 156:
```suggestion
        lambda row: match_nuc_to_aas(row['nuc_change'], row['aa_change']), pl.String
```
I'm not sure if this is a row, but it is a bit better than x

scripts/merge_defining_mutations.py
JonasKellerer: In the expected output there are no longer the `X` symbols, but weh I run the test, I get it in the output.

tests/test_merge_defining_mutations.py
```diff
@@ -0,0 +1,119 @@
...
+        "S:S375F",
+        "S:T376A",
+    ]
+    assert match_nuc_to_aas(nuc, aas) == 'ORF1a:S135R'
+
+
+def test_match_missed_nuc_to_aas():
+    nuc = "X21766-"
+    aas = ['S:H69-']
+    # TODO: this example is also classified differently in corn vs emma (wuhan vs pango_parent)
```
JonasKellerer, line 56: Same as above: Add a ticket.

tests/test_merge_defining_mutations.py
```diff
@@ -0,0 +1,119 @@
...
+    assert match_nuc_to_aas(nuc, aas) == 'S:H69-'
+
+
+def test_match_nuc_to_multiple_aas():
+    nuc = "C28312T"
+    aas = ['N:P13L', 'ORF9b:P10F']
+    # TODO: need to clarify how we incorporate multiple amino acids
```
JonasKellerer, line 63: Same as above: Add a ticket.

tests/test_merge_defining_mutations.py
```diff
@@ -0,0 +1,119 @@
...
+    assert match_nuc_to_aas(nuc, aas) == ['N:P13L', 'ORF9b:P10F']
```
JonasKellerer, line 64: I think this testcase contradicts `test_match_missed_nuc_to_aas`, but maybe you meant this in the comment.

tests/test_merge_defining_mutations.py
JonasKellerer, line 1: I needed to add `pytest` and `pyarrow` to the requirements. But I have found this also in https://github.com/hodcroftlab/covariants/pull/572

tests/test_merge_defining_mutations.py
```diff
@@ -0,0 +1,119 @@
+import json
+import os
+
+from scripts.merge_defining_mutations import process_cornelius_file, process_emma_file, split_nuc, nuc_to_gene, \
+    split_aa, match_nuc_to_aas, main
+
+
+def test_split_nuc():
```
JonasKellerer, line 8: I'm not sure how verbose we want to test, but we do not catch anything that is of the wrong format. So "ABC" is maybe also transformed, but I don't know :)

tests/test_merge_defining_mutations.py
```diff
@@ -0,0 +1,119 @@
...
+    nuc1 = "T670G"
+    nuc2 = "C11289-"
+    nuc3 = "G10447A"
+    nuc4 = "G4-"
+    assert split_nuc(nuc1) == ('T', 670, 'G')
+    assert split_nuc(nuc2) == ('C', 11289, '-')
+    assert split_nuc(nuc3) == ('G', 10447, 'A')
+    assert split_nuc(nuc4) == ('G', 4, '-')
```
JonasKellerer, line 16: I think we can parameterize these tests.

tests/test_merge_defining_mutations.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def test_nuc_to_gene():
+    nuc1 = 670
+    nuc2 = 21636
+    nuc3 = 28312
+    assert nuc_to_gene(nuc1) == ['ORF1a']
+    assert nuc_to_gene(nuc2) == ['S']
+    assert nuc_to_gene(nuc3) == ['N', 'ORF9b']
```
JonasKellerer, line 25: Same as above.

tests/test_merge_defining_mutations.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def test_split_aa():
+    aa1 = 'ORF9b:P10S'
+    aa2 = 'S:P681H'
+    aa3 = 'ORF9b:N28-'
+    assert split_aa(aa1) == ('ORF9b', 'P', 10, 'S')
+    assert split_aa(aa2) == ('S', 'P', 681, 'H')
+    assert split_aa(aa3) == ('ORF9b', 'N', 28, '-')
```
JonasKellerer, line 34: Same as above.

tests/test_merge_defining_mutations.py
```diff
@@ -0,0 +1,119 @@
...
+                                'unaliased',
+                                'parent',
+                                'children',
+                                'nextstrain_clade',
+                                'designation_date']
+    assert len(corn) == 258
+    assert corn.columns == ['lineage', 'nextstrain_clade', 'nuc_change', 'aa_change', 'relative_to']
+
+
+def test_main():
```
JonasKellerer, line 86: This test fails for me.

tests/data/defining_mutations/cornelius/cornelius_edge_cases.json
```diff
@@ -0,0 +1,518 @@
+{
+  "B": {
```
JonasKellerer, line 2: Could we name these edge cases somehow to describe what they should test?

scripts/merge_defining_mutations.py
```diff
@@ -0,0 +1,355 @@
...
+
+
+def remove_empty_strings(y):
+    if len(y) == 1:
+        if y[0] == '':
+            return []
+    return y
+
+
+def process_cornelius_file(path):
```
JonasKellerer, line 74:
```suggestion
def process_cornelius_file(path: str):
```

JonasKellerer (changes requested):
Great work. We are a huge step further.

Most of my comments are regarding refactoring. It feels this is more script style. But it could benefit from some functions to make the intention clearer. Also some more type hints would be nice.

Some of my points are adressed by #572. Feel free to ignore them, but I would then suggest that we merge the other PR as fast as possible, so the build works.

JonasKellerer:
I just noticed something: We in the format of the defining mutations, we have `aa_change` and `nuc_change`. Speaking from the domain language then these would be `aa_mutations` and `nuc_mutations`.

JonasKellerer:
Also what do you think about this structure of the json file? This would make it easier to add more references later on. It also gets rid of the `mutations_relative_to`.

```json
"mutations": [
    {
      "reference": "wuhan",
      "coding_mutations": [
        {
          "aa_mutation": {
            "gene": "ORF1a",
            "ref": "S",
            "pos": 135,
            "alt": "R"
          },
          "nuc_mutations": [
            {
              "ref": "T",
              "pos": 670,
              "alt": "G"
            }
          ]
        }
      ],
      "silent_mutations": [
        {
          "nuc_mutation": {
            "ref": "C",
            "pos": 241,
            "alt": "T"
          }
        }
      ]
    }
  ]
```

JonasKellerer:
I just added a ticket, to keep track of the link between the defining mutations and the clusters.json: https://github.com/hodcroftlab/covariants/issues/580

tests/test_merge_defining_mutations.py
AdvancedCodingMonkey: Exactly, I hope it is clearer now

tests/test_merge_defining_mutations.py
AdvancedCodingMonkey: I think we are ok for the moment :)

AdvancedCodingMonkey (commented):
Tried to address your points, let me know if anything is still missing.

scripts/merge_defining_mutations.py
```diff
@@ -0,0 +1,355 @@
...
+def aa_string_to_object(aa):
+    gene, ref, pos, alt = split_aa(aa)
+    return {'gene': gene, 'ref': ref, 'pos': pos, 'alt': alt}
+
+
+def remove_empty_strings(y):
+    if len(y) == 1:
+        if y[0] == '':
+            return []
+    return y
```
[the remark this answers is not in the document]
JonasKellerer: Okay. I think we should name it differently, so it is clear that it only replaces empty strings in the arrays.

Also a nitpick: I think 
```
if len(y) == 1 and y[0] == '':
```
will do the same.

scripts/merge_defining_mutations.py
JonasKellerer: Thank you. Much better now :)

tests/test_merge_defining_mutations.py
JonasKellerer: Very nice.

tests/test_merge_defining_mutations.py
JonasKellerer: Hm... I'm not sure what changed, but now it works :)

scripts/merge_defining_mutations.py
```diff
@@ -0,0 +1,355 @@
...
+    'ORF6': [27202, 27387],
+    'ORF7a': [27394, 27759],
+    'ORF7b': [27756, 27887],
+    'ORF8': [27894, 28259],
+    'ORF9b': [28284, 28577],
+    'S': [21563, 25384]
+}
+
+
+def split_nuc(nuc: str) -> tuple[str, int, str]:
```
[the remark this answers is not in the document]
AdvancedCodingMonkey: Added data classes now, thanks for the push

scripts/merge_defining_mutations.py
```diff
@@ -0,0 +1,390 @@
...
+    'ORF7a': [27394, 27759],
+    'ORF7b': [27756, 27887],
+    'ORF8': [27894, 28259],
+    'ORF9b': [28284, 28577],
+    'S': [21563, 25384]
+}
+
+
+@dataclass
+class Mutation:
```
JonasKellerer, line 27: Nitpick: This is only a mutation for covid. I general, all mutations have a coresponding segment. But since we only have covid, I think this is fine. And it is much easier this way. 

So the structure would be:
Mutation(segment, from, position, to)
NucleotideMutation(segment=null,from, position, to) : Mutation
AminoAcidMutation(segment=gene,from,position,to): Mutation

JonasKellerer (approved):
Very nice. This brings us much further. Thank you also for all the refactorings.

JonasKellerer:
Remember squash before merge :)

scripts/merge_defining_mutations.py
```diff
@@ -0,0 +1,390 @@
...
+class AminoAcidMutation(Mutation):
+    gene: str
+
+    @classmethod
+    def parse_amino_acid_string(cls, amino_acid_mutations_string: str) -> "AminoAcidMutation":
+        gene, aa =amino_acid_mutations_string.split(':')
+        mut = cls.parse_mutation_string(aa)
+        return AminoAcidMutation(mut.symbol_from, mut.position, mut.symbol_to, gene)
+
+    def to_string(self) -> str:
```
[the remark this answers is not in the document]
AdvancedCodingMonkey: Yes, we could use `__str__`. I like having a bit of a difference here though, since `__str__` gives you a bit more information about the object, like the class name and the field names. Naming it code seems helpful.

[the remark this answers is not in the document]
AdvancedCodingMonkey: I am going to do it on the other branch though, then I don't have to rebase :)

PR #572, opened 2025-04-01
488 add github action

AdvancedCodingMonkey:
Add a github action that pulls defining mutations data from https://github.com/corneliusroemer/pango-sequences/blob/main/data/pango-consensus-sequences_summary.json and merges with the manually curated files.


.github/workflows/defining_mutations.yml
```diff
@@ -0,0 +1,54 @@
...
+      - name: "install python dependencies"
+        run: |
+          pip install -r requirements.txt
+
+      - name: "download defining mutations"
+        run: |
+          mkdir -p data
+          curl -o data/cornelius.json https://raw.githubusercontent.com/corneliusroemer/pango-sequences/refs/heads/main/data/pango-consensus-sequences_summary.json
+
+      - name: "merge files and update web data"
```
JonasKellerer, line 44: Should we make sure, that the tests pass before we run the calculation?

scripts/merge_defining_mutations.py
```diff
@@ -38,13 +38,14 @@ def nuc_to_gene(nuc_location: int):
...
-def match_nuc_to_aas(nuc, aas):
+def match_nuc_to_aas(nuc, aas, verbose=False):
     if not nuc:
         return
     nuc_old, nuc_location, nuc_new = split_nuc(nuc)
     genes = nuc_to_gene(nuc_location)
     if len(genes) == 0:
-        print(f'WARNING: nuc {nuc} cannot be matched to gene region')
+        if verbose:
+            print(f'WARNING: nuc {nuc} cannot be matched to gene region')
```
JonasKellerer, line 48: I think we should use logging now, before we add a verbose flag:
```
import logging
...
logging.info(f'WARNING: nuc {nuc} cannot be matched to gene region')
```
The user can the set the log level.

scripts/merge_defining_mutations.py
```diff
@@ -38,13 +38,14 @@ def nuc_to_gene(nuc_location: int):
     return genes
 
 
-def match_nuc_to_aas(nuc, aas):
+def match_nuc_to_aas(nuc, aas, verbose=False):
```
JonasKellerer, line 41: With the logger, we no longer need this boolean flag.

scripts/merge_defining_mutations.py
```diff
@@ -193,7 +194,12 @@ def process_emma_file(path):
     filename = path.split('/')[-1]
     lineage = filename.split('.')[0]
 
-    data = pl.read_csv(path, separator='\t')
+    data = (
+        pl.read_csv(path, separator='\t')
+        .with_columns(
+            pl.col('nuc_change').str.strip_chars(),
+            pl.col('aa_change').str.strip_chars())
```
JonasKellerer, line 201: Could you explain why this change was necessary?

scripts/merge_defining_mutations.py
```diff
@@ -224,6 +230,12 @@ def check_relative_to_column(df):
         raise ValueError(f'Could not assign reference point for mutations {df.filter(pl.col("relative_to").is_null())}')
 
 
+def sort_mutations(x):
```
JonasKellerer, line 233: A type hint would be nice here and a better name for the variable :)

scripts/merge_defining_mutations.py
```diff
@@ -266,18 +280,29 @@ def merge_file_dfs(emma_df: pl.DataFrame, corn_df: pl.DataFrame):
...
     )
 
-    typed = coalesced.with_columns(
-        mutation_type=pl.when(pl.col('aa_change').is_not_null()).then(pl.lit('coding')).otherwise(pl.lit('silent'))
-    ).select('lineage', 'nextstrain_clade', 'relative_to', 'mutation_type', 'aa_change', 'nuc_change', 'notes')
+    typed = (
+        coalesced
+        .with_columns(
+            mutation_type=pl.when(pl.col('aa_change').is_not_null()).then(pl.lit('coding')).otherwise(pl.lit('silent')))
+    )
```
JonasKellerer, line 287: We could think about a dedicated formatter for our files. Not this PR, but I think it would be nice to have a standart. I'm not sure what this is, but in one of my side projects i used `black`. https://github.com/psf/black. We could then also add it to the ci, to make sure new files have the style.

scripts/merge_defining_mutations.py
```diff
@@ -337,17 +378,18 @@ def reformat_df_to_dicts(merged, lineages):
     return output_dicts
 
 
-def main(emma_dir='../tests/data/defining_mutations/emma', corn_dir='../tests/data/defining_mutations/cornelius',
-         output_dir='../tests/data/defining_mutations/output'):
+def main(emma_dir='defining_mutations', corn_dir='data',
+         output_dir='web/public/data/definingMutations', write_output=True):
```
JonasKellerer, line 382: I think it would be nice to set these varibales from outside and not rely on the default values here.

We could think about `argparse` which is easy enough to use.

JonasKellerer: When is the case that we do not want output? We could add a new method which is then used with an additional flag in the main `--validateFiles` or `--dryRun`.

tests/test_merge_defining_mutations.py
```diff
@@ -65,13 +70,13 @@ def test_match_nuc_to_multiple_aas():
 
 
 def test_process_emma_file():
-    emma = process_emma_file('data/defining_mutations/emma/22E.Omicron.tsv')
+    emma = process_emma_file('tests/data/defining_mutations/emma/22E.Omicron.tsv')
```
JonasKellerer, line 73: Do we want to define some constant, so that we do not need to write `tests/data` all the time? the folder could change at some point.

scripts/merge_defining_mutations.py
```diff
@@ -266,18 +280,29 @@ def merge_file_dfs(emma_df: pl.DataFrame, corn_df: pl.DataFrame):
...
+    typed = (
+        coalesced
+        .with_columns(
+            mutation_type=pl.when(pl.col('aa_change').is_not_null()).then(pl.lit('coding')).otherwise(pl.lit('silent')))
+    )
 
-    coding_grouped_by_aa = (
+    position_sorted = (
         typed
+        .with_columns(nuc_position=pl.col('nuc_change').str.tail(-1).str.head(-1).cast(pl.Int64))
```
JonasKellerer, line 291: Is there a way to store the nuc_change as an object in the dataframe? It would then be easier to extract the position from it. This feels a bit shaky.

scripts/merge_defining_mutations.py
```diff
@@ -287,12 +312,28 @@ def merge_file_dfs(emma_df: pl.DataFrame, corn_df: pl.DataFrame):
 
 
 def reformat_df_to_dicts(merged, lineages):
-    aggregate_mutations = (
-        merged
+    valid_mutations = merged.filter(pl.col('nuc_change').is_not_null())
+    no_mutations = merged.filter(pl.col('nuc_change').is_null())
```
JonasKellerer, line 316: I don't understand this change. Do we want to include invalid mutations in the output?

scripts/merge_defining_mutations.py
```diff
@@ -241,15 +253,17 @@ def import_file_dfs(emma_dir, corn_dir):
...
-    corn_raw = corn_df.with_columns(nuc_change_raw=pl.col('nuc_change').str.slice(1))
-    combined = (corn_raw
-                .join(emma_raw,
-                      on=['lineage', 'nextstrain_clade', 'nuc_change_raw', 'relative_to'],
-                      how='full',
-                      suffix='_emma',
-                      coalesce=True)
-                .drop('nuc_change_raw'))
+    emma_raw = emma_df.with_columns(nuc_change_raw=pl.col('nuc_change').str.tail(-1).str.head(-1))
+    corn_raw = corn_df.with_columns(nuc_change_raw=pl.col('nuc_change').str.tail(-1).str.head(-1))
```
JonasKellerer, line 257: I'm not sure that I understand what happens here. Does this extract the position? If so, I refer to my other comment, if we could store the change as an object and get the position from there.

JonasKellerer (changes requested):
Nice. Good to have automated tests now.

I think it would be good to split this PR into two parts, one which adds the workflow and the changes to the python files. And one that adds the mutations. But we can do this during merge.

My comments are mostly concerning refactorings. 

Maybe not this PR, but I would advice to use logging, formatting and argparse to run our scripts.

scripts/merge_defining_mutations.py
AdvancedCodingMonkey: Removed after discussion with Emma

scripts/merge_defining_mutations.py
AdvancedCodingMonkey: Whoops, forgot to delete the function, it is not actually used.

scripts/merge_defining_mutations.py
AdvancedCodingMonkey: I like defaults because they show the intention and turn positional arguments into keywords. How about we do both? Defaults that will be used when calling the main function and command line args when running the script?

scripts/merge_defining_mutations.py
AdvancedCodingMonkey: Aaaand its done.

scripts/merge_defining_mutations.py
JonasKellerer: Good idea. 

Just a question: Why are the roots of the `output_dir` and `auto_generated_data_dir` different in the default value? One time it is the root of the repo and one time the test folder?

scripts/merge_defining_mutations.py
JonasKellerer: Nice.

tests/data/defining_mutations/auto_generated_edge_cases/auto_generated.json
```diff
@@ -0,0 +1,790 @@
...
+    "nucSubstitutionsReverted": [],
+    "aaSubstitutionsReverted": [],
+    "nucDeletionsReverted": [],
+    "aaDeletionsReverted": [],
+    "frameShifts": [
+      ""
+    ],
+    "designationDate": "2022-08-25"
+  },
+  "LP.5": {
```
JonasKellerer, line 518: This case has no name assigned to it. :)

JonasKellerer (approved):
Looks great. 

Not something for this PR, but we could think about having smaller files to test, so the tested thing is directly visible. But maybe we need those large files.

scripts/merge_defining_mutations.py
JonasKellerer: Ah. So the testing folder `data` is not this data folder?

AdvancedCodingMonkey: No, it is a data folder at the root that the backend user has to create (also gets created with the github action).

PR #583, opened 2025-04-04
refactor: use nextjs layout to define general layout of pages

JonasKellerer:
Moves the layout out of the "pages" files and moves them into a layout, to make the concept clear: We have a basic layout with header and footer. In this we have narrow and full width pages. We then render all the content in these containers.

Ref: https://nextjs.org/docs/pages/building-your-application/routing/pages-and-layouts#layout-pattern

Also removes some of the `<Row>` and `<Col>` tags, since a basic div does the same. They are only usefull, when having more than one column.


web/src/components/Layout/WithHeadline.tsx
```diff
@@ -0,0 +1,14 @@
+import React, { PropsWithChildren } from 'react'
+import { PageHeading } from '../Common/PageHeading'
+import { useTranslationSafe } from 'src/helpers/useTranslationSafe'
+
+export function WithHeadline({ children, title }: PropsWithChildren<{ title: string }>) {
```
AdvancedCodingMonkey, line 5: This is nice

web/src/components/Layout/PageSizes/NarrowPage.tsx
```diff
@@ -0,0 +1,10 @@
+import React, { PropsWithChildren } from 'react'
+import { Layout } from 'src/components/Layout/Base/Layout'
+
+export function NarrowPage({ children }: PropsWithChildren) {
+  return (
+    <Layout>
+      <div className="container-xl">{children}</div>
```
AdvancedCodingMonkey, line 7: xl = narrow :laughing:

web/src/components/Home/HomePage.tsx
```diff
@@ -1,32 +1,14 @@
 import React from 'react'
-
-import { Col, Row } from 'reactstrap'
 import { ClusterButtonPanelLayout } from 'src/components/ClusterButtonPanel/ClusterButtonPanelLayout'
-import { NarrowPageContainer } from 'src/components/Common/ClusterSidebarLayout'
 import { Editable } from 'src/components/Common/Editable'
-import { Layout } from 'src/components/Layout/Layout'
 import { MdxContent } from 'src/i18n/getMdxContent'
 
 export function HomePage() {
```
AdvancedCodingMonkey, line 6: Nice to get rid of all this code. The naming `...Page` feels even less accurate than before though. No need to refactor though.

web/src/components/CountryDistribution/CountryDistributionPage.tsx
```diff
@@ -29,56 +27,23 @@ import { clusterSidebarCollapsedAtoms, countriesSidebarCollapsedAtoms } from 'sr
...
   }, [contentFilename])
 
   return (
-    <>
-      <Row className={'gx-0'}>
-        <Col>
-          <RegionSwitcher />
-        </Col>
-      </Row>
+    <div>
```
AdvancedCodingMonkey, line 36: Why not use `<>`?

web/src/components/ClusterDistribution/ClusterDistributionPage.tsx
```diff
@@ -25,35 +23,17 @@ export function ClusterDistributionPage() {
...
-
-      <Row className={'gx-0'}>
-        <Col className="pb-10">
-          <Editable githubUrl="blob/master/scripts" text={t('View data generation scripts')}>
-            <ClusterDistributionPlotSection />
-          </Editable>
-        </Col>
-      </Row>
-    </Layout>
+    <div>
```
AdvancedCodingMonkey, line 26: Why not use `<>`?

web/src/components/Cases/CasesPage.tsx
```diff
@@ -22,35 +20,17 @@ export function CasesPage() {
   const { t } = useTranslationSafe()
 
   return (
-    <Layout wide>
-      <Row className={'gx-0'}>
-        <Col>
-          <PageHeading>{t('Estimated Cases by Variant')}</PageHeading>
-        </Col>
-      </Row>
+    <div>
```
AdvancedCodingMonkey, line 23: Why not use `<>`?

web/src/components/SharedMutations/SharedMutationsPage.tsx
```diff
@@ -1,66 +1,31 @@
...
-                    <SharedMutationsTable />
-                  </Suspense>
-                </ErrorBoundary>
-              </SharedMutationsWrapper>
-            </Editable>
-          </Col>
-        </Row>
-      </SharedMutationsPageContainer>
-    </Layout>
+    <div>
```
AdvancedCodingMonkey, line 15: Why not use `<>`?

web/src/components/Variants/VariantsPage.tsx
```diff
@@ -201,3 +162,37 @@ export function VariantsPageContent({ currentCluster }: { currentCluster: Cluste
...
+    flex: 1 0;
+  }
+`
+
+const FlexGrowing = styled.div`
+  display: flex;
+  flex: 1 0;
+`
+
+const EditableClusterContent = styled(Editable)``
```
AdvancedCodingMonkey, line 189: Let's remove this "empty" styled component. It would also be nice to trim the scope of `EditableClusterContent` so that it only wraps the md file and the link to the nextstrain build. But I am guessing fixing the layout consequences will be annoying, so let's leave that for a later refactoring.

web/src/pages/_app.tsx
```diff
@@ -69,9 +79,7 @@ function MyApp({ Component, pageProps }: AppProps) {
             <Plausible domain={DOMAIN_STRIPPED} />
             <SeoApp />
             <QueryClientProvider client={queryClient}>
-              <Suspense fallback={LOADING}>
-                <Component {...pageProps} />
-              </Suspense>
+              <Suspense fallback={LOADING}>{getLayout(<Component {...pageProps} />)}</Suspense>
```
AdvancedCodingMonkey, line 82: Reads a bit weird to put the component back into its own getLayout function but actually quite nice.

AdvancedCodingMonkey (approved):
Very nice!

web/src/components/Layout/Base/NavigationBar.tsx
```diff
@@ -60,7 +60,7 @@ export function NavigationBar() {
 
 function Brand() {
   return (
-    <Link href="/">
+    <Link href="/public">
```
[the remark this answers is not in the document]
JonasKellerer: Good find. It was a refactoring of intelliJ and this should not have been changed.

web/src/pages/variants/[...clusterName].tsx
```diff
@@ -33,4 +35,10 @@ export async function getStaticPaths(_0: GetStaticPathsContext): Promise<GetStat
   }
 }
 
-export { VariantsPage as default } from 'src/components/Variants/VariantsPage'
+export default function Variants({ clusterName }: { clusterName: string }) {
+  return <VariantsPage clusterName={clusterName} />
+}
+
+Variants.getLayout = function getLayout(page: ReactElement) {
+  return <NarrowPage>{page}</NarrowPage>
```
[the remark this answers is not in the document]
Comment (JonasKellerer) [reply web/src/pages/variants/[...clusterName].tsx:43]: True. I tried it and it was quite hard. The problem was, that we need to calculate and fetch the cluster from the clusterName. And this cluster is then not passed to the component, since we do not have a method to pass this from the `getLayout` function to the Page. Also we have some logic tied to the clusterName, namely `useDeriveCurrentClusterNameFromUrl`, where we change the url, depending on the state of the atoms.

I created https://github.com/hodcroftlab/covariants/issues/584 , which addresses a similar issue, where we need the title of the page in the layout. I think we can try there again.

web/src/components/Layout/PageSizes/NarrowPage.tsx
JonasKellerer: 😆 Maybe I rename it to `LimitedWidthPage`, to be in line with the `FullWidthPage`.

web/src/components/CountryDistribution/CountryDistributionPage.tsx
JonasKellerer: A good question: When using the fraction we are no longer sure that our component is the only element in the surrounding element. If that would be the case, then some formatting of the surrounding element will influence our page. For example `d-flex flex-wrap` in the surrounding element, would also wrap our elements, so the `RegionSwitcher` could be next to the `IntroContent`.

web/src/components/Variants/VariantsPage.tsx
JonasKellerer: Good find. I wanted to ged rid of the styled components one by one in a separate PR. This PR was growing to be very large.

web/src/pages/_app.tsx
JonasKellerer: I agree. But its nice to have the layout function directly at the component.

web/src/components/Home/HomePage.tsx
JonasKellerer: I changed the names of the components in the pages folder to page and the components to the name without "Page"

web/src/components/Cases/CasesPage.tsx
JonasKellerer: Same answer as above

web/src/components/SharedMutations/SharedMutationsPage.tsx
JonasKellerer: Same answer as above

web/src/pages/_app.tsx
AdvancedCodingMonkey, line 82: As I am looking at it a second time, I feel we are not using this pattern as intended. It would not make a difference right now if each page was returned with the layout instead of having a function construct the layout. We would actually have to re-use the pages themselves to utilize the `getLayout` pattern. I actually don't see us re-using the pages as components, since we already have components pulled out from the pages that we can re-use should it become necessary. To keep things simple, we could either move the components from their own files into the pages files or have the pages files return the layout directly rather than providing a `getLayout` function. I would vote for the latter but maybe you have a different take.

JonasKellerer: As discussed: 
We get rid of the `getLayout` pattern and move the layout to the pages directly. There is no direct benefit to have this extra function, besides providing a guidline, where to put the layout. We decided to put the generic page layout directly in the pages files. This also allows us also to set the title of the page from the request params.

PR #614, opened 2025-05-20
feat-612: fix column width of defining mutations table

JonasKellerer:
-  adds the option to specify for each table, that each of the columns has the same width.
- fixes minor spelling
- makes the table "responsive" meaning, that when the minimal width (length of the title) is reached, then there appears a scroll bar at the bottom

grafik


AdvancedCodingMonkey (approved):
Nice

Issue #409, opened 2024-12-02
Allow zooming


AdvancedCodingMonkey:
- Allow zooming (there's some work on this https://github.com/hodcroftlab/covariants/pull/344)

AdvancedCodingMonkey:
See also https://recharts.org/en-US/examples/HighlightAndZoomLineChart for an example

JonasKellerer:
More info on this ticket:
- add to all plot cards (per country, variants, cases)
- add reset button to plot
- on per country page zoom in every plot at once

PR #619, opened 2025-05-22
feat-409: add zooming to plots

JonasKellerer:
Resolves: #409

Adds zooming to the plots on the perVariant, perCounrty and cases pages. The zooming window is stored between page changes. It is applied to all plots on the respective pages.

grafik

### DIFF
--- a/web/src/state/Params.ts
+++ b/web/src/state/Params.ts
@@ -1,4 +1,4 @@
-import { atom, selector } from 'recoil'
+import { atom, atomFamily, selector, selectorFamily } from 'recoil'
 import { DateTime, Interval } from 'luxon'
 import { fetchParams, GlobalParams } from 'src/io/getParams'
 import { dateStringToSeconds } from 'src/helpers/format'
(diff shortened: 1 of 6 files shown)


web/src/state/Params.ts
```diff
@@ -18,37 +18,48 @@ const INVALID_PARAMS = 'Invalid date params in params.json or invalid split poin
...
-    const months = interval.splitBy({ months: 1 })
-    if (months.length === 0 || months.some((i) => !i.isValid)) {
-      throw new Error(INVALID_PARAMS)
-    }
-    const validMonths = months as Interval<true>[]
-
-    return validMonths.map((d) => d.start.toSeconds())
-  },
+  get:
+    (param) =>
```
AdvancedCodingMonkey, line 47: See my comment above

AdvancedCodingMonkey (approved):
Works like a charm, implemented at lightning speed. If one leaves the chart window while holding down the mouse, there is a bit of a glitch happening (when the mouse comes back and clicks, only 1 datapoint is selected but I find this acceptable.

PR #622, opened 2025-05-26
feat-615: add who label

### LINKED ISSUE #615: Add WHO info for clades

AdvancedCodingMonkey:
Resolves #615 
Adds the who label from the nextclade to tree to a clade's defining mutations page.

### DIFF
--- a/web/src/components/DefiningMutations/SelectReferenceDropdown.tsx
+++ b/web/src/components/DefiningMutations/SelectReferenceDropdown.tsx
@@ -1,10 +1,7 @@
 import React, { useCallback, useId, useMemo } from 'react'
+import _ from 'lodash'
 import { useTranslationSafe } from 'src/helpers/useTranslationSafe'
-import {
-  DropdownOption,
-  stringsToOptionsWithLabels,
-  stringToOptionWithLabel,
-} from 'src/components/Common/DropdownOption'
+import { DropdownOption, stringsToOptions, stringToOption } from 'src/components/Common/DropdownOption'
 import { Dropdown } from 'src/components/Common/Dropdown'
 
 const style = { maxWidth: '300px' }
@@ -21,7 +18,7 @@ export function SelectReferenceDropdown({
   const { t } = useTranslationSafe()
   const id = useId()
 
-  const options = useMemo(() => stringsToOptionsWithLabels(referenceSequences, toSentenceCase), [referenceSequences])
+  const options = useMemo(() => stringsToOptions(referenceSequences), [referenceSequences])
 
   const onChange = useCallback(
     (newValue: DropdownOption<string>) => {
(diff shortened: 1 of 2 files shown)


web/src/components/DefiningMutations/SelectReferenceDropdown.tsx
```diff
@@ -33,11 +37,20 @@ export function SelectReferenceDropdown({
...
     </div>
   )
 }
+
+const toSentenceCase = (str: string) => {
+  const s = str?.match(/[A-Z]{2,}(?=[A-Z][a-z]+[0-9]*|\b)|[A-Z]?[a-z]+[0-9]*|[A-Z]|[0-9]+/g)?.join(' ')
+  if (s === undefined) {
+    return str
+  }
+  return s.slice(0, 1).toUpperCase() + s.slice(1)
```
JonasKellerer, line 55:
```suggestion
return _.capitalize(_.startCase(str))
```
I think this does the same and is easier to follow.

web/src/components/DefiningMutations/SelectReferenceDropdown.tsx
```diff
@@ -33,11 +37,20 @@ export function SelectReferenceDropdown({
       </label>
       <Dropdown
         identifier={id}
+        className={'width-5xs'}
         options={options}
-        value={stringToOption(selectedSequence)}
+        value={stringToOptionWithLabel(selectedSequence, toSentenceCase)}
```
JonasKellerer, line 42: Do we really need to format the string in the option and in the value? I think this would also do the trick:
```
formatOptionLabel={(option) => toSentenceCase(option.value)}
```
We could also remove the value from the prop here. We don't need to make the input controlled. This way we have less the pass to the component. We only have one place where we want to change the value, and that is with this dropdown.

web/src/styles/_utilities.scss
```diff
@@ -0,0 +1,16 @@
...
+    $utilities,
+    (
+      "width": (
+        property: width,
+        values: (
+          5xs: 12rem,
+        )
+      )
+    )
+);
```
JonasKellerer: Nice :) 
I would have expected something like `w-48`, but this is because I have never used the tailwind utility classes in "sizes" :)

JonasKellerer (approved):
Looks good. Only some minor refactorings.

web/src/components/DefiningMutations/SelectReferenceDropdown.tsx
AdvancedCodingMonkey: I don't quite understand what you mean by "removing the value from the prop" and find the Dropdown API a bit confusing. I played with the formatOptionLabel but somehow did it wrong. Now it works, thanks.

